PR #2, opened 2015-07-01
Passed props should override defaults

tikotzky:
The way it is now if you try to set the `display` property of an `Image` it is overwritten by the defaults.

This PR moves the styles from props to the end of the assign call they will always override the defaults.


kenwheeler:
Thanks!

tikotzky:
Thank **you** for the awesome lib!

PR #15, opened 2015-07-02
Correct typo in imports

mdamien:
This typo makes the import not work on my machine

Some versions of what I use:
```
node v0.10.25
npm 1.3.10
webpack 1.10.1
```


kenwheeler:
Great catch, thanks!

PR #16, opened 2015-07-02
Adds Surge as a development dependency for deployment

kennethormandy:
If you don’t already have Surge installed with `npm install -g surge`, the `npm run deploy` command will fail. This fixes the issue by adding `surge` as a `devDependency` in the `package.json`.


kenwheeler:
Thanks!

PR #29, opened 2015-07-07
overview mode (fix #17)

### LINKED ISSUE #17: Add Overview Mode
Basically, the same as implemented in reveal.js.

mathieudutour:


mathieudutour:
it's quite a big refactoring, happy to hear your thoughts

kenwheeler:
Can't wait to check it out, you're a machine!

src/appear.jsx
```diff
@@ -32,8 +32,7 @@ const Appear = React.createClass({
     this.context.flux.stores.SlideStore.unlisten(this._storeChange);
   },
   _storeChange(state) {
-    const slide = "slide" in this.context.router.state.params ?
-      this.context.router.state.params.slide : 0;
+    const slide = this.context.slide;
```
kenwheeler: been meaning to do this. this is great.

src/deck.jsx
```diff
@@ -49,37 +50,56 @@ class Deck extends React.Component {
...
       this._nextSlide();
     }
+    if (event.keyCode === 79) { // o
+      this._toggleOverviewMode();
+    }
+    if (event.keyCode === 80) { // o
+      this._togglePresenterMode();
+    }
+  }
+  _toggleOverviewMode() {
```
kenwheeler: We should probably clear the flux stores when going into these modes. Sometime you have to hit the right arrow double as many times to get the fragments to start appearing.

src/slide.jsx
```diff
@@ -33,10 +36,12 @@ const Slide = React.createClass({
   setZoom() {
     const content = React.findDOMNode(this.refs.content);
     const zoom = (content.offsetWidth / config.width);
-    const contentScale = (content.parentNode.offsetHeight / config.height);
+    const contentScaleY = (content.parentNode.offsetHeight / config.height);
+    const contentScaleX = (content.parentNode.offsetWidth / config.width);
```
kenwheeler: We shouldn't do this. Open up the presentation in regular mode and scale to mobile size, you'll see what I mean.

kenwheeler:
Just added a couple notes, other than that stuff, this looks great.

src/slide.jsx
mathieudutour: agreed but there are some issues in the presenter mode and overview mode: the preview doesn't match the actual slide. Don't really know what to do. Need to dig in more

src/deck.jsx
mathieudutour: Not sure how to do it. What we really want is not clearing totally the store but bypassing the `Appear` fragment: if i'm in a middle of a slide with some Appears, going into overview mode and back again should keep the Appears already revealed

kenwheeler: I'll do it. I think the issue is that they are being indexed in flux by their internal react ID, which is different in the modes. If I can ID them by `slide + order of appearance` it should match up.

src/slide.jsx
kenwheeler: In the current presenter mode, on desktop at least, it looks ok. Mobile is questionable, but generally useable. I see the difficulty for overview though, because you can't really flexbox that layout.

src/deck.jsx
kenwheeler: This is a tricky one. Debating enforcing a ref on fragments to help with non instance based identification. Thoughts?

mathieudutour: yeah might do the trick

src/slide.jsx
mathieudutour: i think i found it

kenwheeler:
This is looking good, but we have to figure out how to keep the appears in sync.

src/deck.jsx
kenwheeler: Ok so requiring a keying prop works magically. But requiring a key sucks, so I'm going to see if I can find any way of doing that without requiring the user to add a key.

mathieudutour: ok so it's actually dead simple.

mathieudutour: it's even keeping the Appear already revealed

kenwheeler: What do you mean?

mathieudutour: test the last commit

mathieudutour: test the last commit

kenwheeler: switch to presenter mode, navigate fragments and switch back

mathieudutour: oh sorry i was speaking about the overview mode

kenwheeler: Indeed it is. But with the new functionality which makes switching back and forth a lot easier I can see the bug being more visible.

mathieudutour: the bug only appear when switching to presenter mode, not when switching to overview view. It's not that you would usually do, maybe there is no need for a quick switch ?

mathieudutour:
Maybe let's merge this one and work on a fix for the duplication of fragments in another PR since this one is quite big already.

PR #30, opened 2015-07-08
add support for page up & page down

tikotzky:
Just gave a talk with one of those presenter remotes and it used page up and page down to move through the slides.

I threw this in before the talk and it went great :smile: 

The same presenter remote worked fine with [reveal.js](https://github.com/hakimel/reveal.js/blob/eff2265bbd6d892ae2316edff748754b4b0c4798/js/reveal.js#L3808-L3811) so i'm assuming its standard.


kenwheeler:
Perfect, thanks dude

PR #32, opened 2015-07-08
allow to extend Slide

mathieudutour:
allow to do `class Slide1 extends Slide` and keep everything related to a slide in its own file.

Unfortunately, i needed to remove react-tween-state as it doesn't support ES6 class.


mathieudutour:
it adds some boilerplate though

kenwheeler:
Not sure I like this level of boilerplate. Does this kill transitions?

mathieudutour:
no no, it uses css transition to do the job

mathieudutour:
my motivation was spliting each slide in its own file. But it's not possible to put everything related to a slide inside it: all the props still need to be defined in the parent file.
Also, if you want to merge the notes PR, being able to define the notes in this slide file is really useful and doesn't' mess with the markup

kenwheeler:
@mathieudutour It's definitely a feature that's been requested a couple times and I can see the value. I wonder if this can be accomplished without the wrapper, like Radium, the class definition , etc... Perhaps it could be accomplished with `raw-loader`. 

I'll merge the notes in a bit, I gotta do it by hand and clean the lint up.

src/deck.jsx
```diff
@@ -3,7 +3,7 @@
 
 import React from "react/addons";
 import assign from "object-assign";
-import cloneWithProps from "react/lib/cloneWithProps";
+const cloneElement = React.cloneElement;
```
kenwheeler: definitely been meaning to get this in too. good stuff.

kenwheeler:
@mathieudutour So using this syntax, or at least separate slide syntax might fix the Appear issue. In a nutshell, we want to pass the slide index down via context. We have two options.

1. Upgrade to 0.14 beta, and use parent based context
2. Use separate slide syntax, where the slide becomes the owner of Appear and allows for us to pass the index down via context.

Thoughts?

kenwheeler:
btw, I would probably pull the tweenState logic right into the slide component, as CSS transitions are jank city.

kenwheeler:
Option 3, confirmed working:

Add `fid` prop to any Appear component, use to resolve duplicates caused by presentation mode.

kenwheeler:
Ok I'm going to close this one because I'm going to do it a bit differently. The new context setup in 0.14 should allow for external slides.

PR #36, opened 2015-07-09
remove fid from Appear

mathieudutour:
I think i found a way to remove the need of fid in Appear. Not really the `react` way but well, it works. Check it out


kenwheeler:
How did I not see this... lol. Awesome dude. That whole setup isn't really the React way, but there really isn't a React way to do this feature, at least until 0.14 is out of beta.

PR #72, opened 2015-09-09
Change keys for Presenter and Overview

jrios:
The o key and p key currently drive the overview and presenter views.

This poses a problem when building an interactive component with a textbox or textarea within a slide. If the o or p key is pressed while in a text input, the slideshow will go into that mode, which may be unexpected behavior to a lot of people.

This pull request requires the alt key to be pressed in conjunction with the p key and the
o to trigger the overview and presenter views.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -35,7 +35,7 @@ Check it out:
 
 ![http://i.imgur.com/H7o2qHI.gif](http://i.imgur.com/H7o2qHI.gif_)
 
-You can toggle the presenter or overview mode by pressing respectively `p` and `o`.
+You can toggle the presenter or overview mode by pressing respectively `alt+p` and `alt+o`.
 
 ## PDF Export
 
--- a/src/deck.jsx
+++ b/src/deck.jsx
@@ -58,10 +58,10 @@ class Deck extends React.Component {
     if (event.keyCode === 39 || event.keyCode === 34) {
       this._nextSlide();
     }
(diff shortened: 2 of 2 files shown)


kenwheeler:
cc @ryan-roemer This should be helpful, want it in the next release?

waywaaard:
You could also just check the target element of the event and prevent triggering the pause and overview when we are typing in a input or textarea.

ryan-roemer:
@jrios thanks for the work@

@kenwheeler I don't have editable code areas, so I'm not affected. But change makes sense. Btw, here's what I do for custom events for my handheld remote: https://github.com/ryan-roemer/surge2015/blob/master/presentation/deck.jsx#L44-L64

```js
// Custom overrides for the deck
class CustomDeck extends Deck {
  _handleKeyPress(e) {
    // Call base method.
    super._handleKeyPress(e);

    // Add some extra key bindings for Satechi remote.
    /*globals window:false*/
    const event = window.event || e;

    const SATECHI_PREV_KEY = 38;
    if (event.keyCode === SATECHI_PREV_KEY) {
      this._prevSlide();
    }

    const SATECHI_NEXT_KEY = 40;
    if (event.keyCode === SATECHI_NEXT_KEY) {
      this._nextSlide();
    }
  }
}
```

jrios:
Thanks for merging! Great project.

kenwheeler:
Thanks for the PR! Sorry it took so long!

jrios:
No problem at all. Thanks for maintaining. Looking forward to using it more.

PR #74, opened 2015-09-19
Adds string hash capability

kenwheeler:
cc @ryan-roemer 

This adds custom hash values via the `id` prop on the `Slide` tag.


ryan-roemer:
@kenwheeler -- LGTM! (But haven't tried it out.. Let me know if I should!)

PR #76, opened 2015-09-22
Update linting dependencies, fix errors

zpao:
I was giving this a try for a presentation and got annoyed by the lint issues showing up my editor so I fixed them.

I left you with with as close as you seemed like you wanted to be to actually following Walmart's config. The biggest deviation was the indent rule to allow case indentation which is what you were doing. I also disabled the line length rules because it doesn't seem like you actually follow it.

And then I filled in the props (mostly just `style`).


zpao:
Actually, have some more coming…

kenwheeler:
Awesome, thanks so much!

PR #80, opened 2015-10-01
Update to work with react 0.14.0-rc1 and react-router 1.0.0-rc1

jimpick:
Not thoroughly tested, but it seems to mostly work.


src/utils/context.jsx
```diff
@@ -26,11 +30,13 @@ const context = function context(Component, params) {
       }
       let slide = 0;
       if (this.props.params && "slide" in this.props.params) {
-        slide = this.props.params.slide;
+        slide = parseInt(this.props.params.slide);
```
AlanFoster: We don't explicitly need it; But can we either use `Number` or pass in our radix here, `parseInt(this.props.params.slide, 10)`

kenwheeler:
Closing pending pull of #90

PR #90, opened 2015-11-03
update dependencies, refactor (ES6)

knowbody:
Mainly updated React Router to 1.0.0-rc4 and React 0.14.2.

Done a bit of refactoring (ES6)


pedrottimark:
While you are at it, should hjs-webpack move from dependencies to devDependencies?

knowbody:
@pedrottimark definitely yes :)

kenwheeler:
@pedrottimark @knowbody what do you guys think about factoring hjs webpack out after this is merged. Moving to a react-transform based build?

kenwheeler:
So, the didLeave errors are fixed and releasing soon https://github.com/facebook/react/issues/4876

src/deck.jsx
```diff
@@ -28,7 +23,7 @@ class Deck extends React.Component {
       lastSlide: null
     };
   }
-  componentDidMount() {
+  componentWillMount() {
```
kenwheeler: Why this change?

src/progress.jsx
```diff
@@ -100,13 +104,13 @@ class Progress extends Base {
 }
 
 Progress.propTypes = {
-  items: React.PropTypes.array,
-  currentSlide: React.PropTypes.number,
-  type: React.PropTypes.oneOf(["pacman", "bar", "number", "none"])
+  items: PropTypes.array,
+  currentslide: PropTypes.number,
```
kenwheeler: Typo?

src/utils/context.jsx
```diff
@@ -26,10 +29,12 @@ const context = function context(Component, params) {
       }
       let slide = 0;
       if (this.props.params && "slide" in this.props.params) {
-        slide = this.props.params.slide;
+        slide = Number(this.props.params.slide);
```
kenwheeler: This value can be a hash string

pedrottimark:
@kenwheeler @knowbody Without any disrespect to Henrik Joretag, unless you expect hjs-webpack to become mainstream, it seems worth the effort to divide webpack settings according to principle of loose coupling and tight cohesion. Combining all the options (for example, transpilers with dev server and hot reloading) makes it harder to understand, not easier, for me at least.

src/progress.jsx
knowbody: yup, thanks :)

src/utils/context.jsx
knowbody: so should I stay as it was?

knowbody:
@kenwheeler updated. And I agree with @pedrottimark about hjs-webpack, not sure why but I find it very confusing, maybe because I was playing with webpack a lot and I don't find it difficult to config, at least not for the things I used it for. Also agreeing with @kenwheeler to do it after this PR is merged.

package.json
```diff
@@ -12,39 +12,42 @@
...
     "normalize.css": "3.0.3",
-    "object-assign": "3.0.0",
-    "radium": "0.13.2",
-    "react": "0.13.3",
-    "react-router": "1.0.0-beta2",
-    "react-tween-state": "0.0.5",
-    "style-loader": "0.12.3",
-    "stylus-loader": "1.2.1",
-    "tween-functions": "1.0.2"
+    "radium": "0.14.2",
```
pedrottimark: @knowbody Can you please confirm that packman progress has animation of moving jaws when a slide changes wIth any of the following versions of radium: 0.13.2, 0.15.0, 0.15.1. I didn't narrow it down any more precise than that.

src/transitions.jsx
```diff
@@ -122,7 +121,7 @@ export default {
 
   },
   componentWillLeave(cb) {
-    const slide = parseInt(this.context.router.state.params.slide) || 0;
+    const slide = parseInt(this.props.currentSlide) || 0;
```
kenwheeler: why is this changed? This messes up slide transition direction. @knowbody

knowbody: in React Router 1.0 there is no more `router` on the `context`

kenwheeler: You'll have to get that slide from elsewhere. location perhaps?

kenwheeler: it looks like slide is on context, so if you add `slide` to contextTypes you can reference `this.context.slide`

knowbody: lol, very good point, give me a minute, fixing it

knowbody:
@kenwheeler should be good now

src/utils/context.jsx
pedrottimark: @knowbody @kenwheeler 1. When I added a slide with a string id, forward navigation did fail with Number(this.props.params.slide) and succeeded when I changed this line back to this.props.params slide. 2. But there are two warnings about the type of slide, so its contextType needs to become PropTypes.oneOfType([PropTypes.number,PropTypes.string]) here and also in src/deck.jsx

knowbody: yeah, I think this is what @kenwheeler was pointing out to me and yeah `oneOfType` is the solution to that

src/deck.jsx
```diff
@@ -275,7 +266,7 @@ class Deck extends React.Component {
   _renderSlide() {
     const slide = this._getSlideIndex();
     const child = this.props.children[slide];
-    return cloneWithProps(child, {
+    return cloneElement(child, {
       key: slide,
```
kenwheeler: Ok this is a fucking hoot. With 0.14, you have to add `children: React.Children.toArray(child.props.children)` here for it to work. Otherwise, TransitionGroup loses its shit and remounts. I have no idea why.

PR #91, opened 2015-11-04
peerDependencies will be deprecated soon.

nicholaswyoung:
To mitigate these errors, we must explicitly include all dependencies.


kenwheeler:
Closing via dep update in the 0.14 branch, thank you

PR #94, opened 2015-11-08
Add presenter query param to presenter link

matthewlehner:
Quick fix on the link link to http://localhost:3000/#/?presenter – didn't have the query param in the `href`, now it does. :smile:


kenwheeler:
Thanks so much

Issue #106, opened 2015-11-19
Pacman progress a solid dot without moving jaws


pedrottimark:
Apparent regression in radium 0.14.2
It works in newer 0.15.0 and 0.15.1 as in original 0.13.2

adamdicarlo:
Seeing this after cloning spectacle-boilerplate yesterday. `npm ls` tells me I have Spectacle 1.0.5 which uses Radium 0.17.1.

Looks like pacman's issue has something to do with this:

`animation:[object Object] 0.12s linear 10 alternate both`

BenDiuguid:
This works on a clean install of [Spectacle Boilerplate](https://github.com/FormidableLabs/spectacle-boilerplate/)

with npm -v : 3.9.5 and node -v : 6.2.2

👍

kenwheeler:
Thanks for confirming, closing this out

Issue #100, opened 2015-11-17
Should Cite support a style prop?


pedrottimark:
Most components support a style prop, although it is undocumented.

For example, here is an excerpt from Quote:

      <span style={[this.context.styles.components.quote, this.getStyles(), this.props.style]}>
        {this.props.children}
      </span>

Quote.propTypes = {
  children: PropTypes.node,
  style: PropTypes.object
};

Cite supports theme and Base style props, but not a style prop:

      <cite style={[this.context.styles.components.cite, this.getStyles()]}>
        - {this.props.children}
      </cite>

Cite.propTypes = {
  children: PropTypes.node
};

pedrottimark:
One zero zero, what an honor. But maybe you want to throw some stale sea weed at me, ha ha?

kenwheeler:
It should

Issue #99, opened 2015-11-17
Should Code and CodePane have (Base) under Element Tags in doc?


pedrottimark:
The descriptions of Code and CodePane under Element Tags omit (Base) but the components call this.getStyles() and in the example deck CodePane has a margin prop.

The only others without (Base) are Appear and S. Those components do not use base props.

kenwheeler:
They should

Issue #98, opened 2015-11-17
Does lineHeight prop apply to non-fit Header and Text?


pedrottimark:
The only reference is in style.text.lineHeight for the contained element that is rendered if fit prop is true.

kenwheeler:
it should

PR #107, opened 2015-11-19
fix #97, #98, #99, #100, #106

### LINKED ISSUE #97: Does false differ from undefined for boolean base style props?
For example, in the default theme, heading styles include fontWeight: "bold" and besides, that is in default formatting of the HTML elements h1, h2, h3, h4, h5, h6.

Should a false value of the bold prop for a Heading component mean this instance is not bold?

In the current logic of getStyles in base.jsx, false has same (non) effect as undefined.

Depending on the intended meaning, the logic might become something like:

if (typeof this.props.italic === "boolean") {
  styles.fontStyle = this.props.italic ? "italic" : "normal";
}
if (typeof this.props.bold === "boolean") {
  styles.fontWeight = this.props.bold ? "bold" : "normal";
}
if (typeof this.props.caps === "boolean") {
  styles.textTransform = this.props.caps ? "uppercase" : "none";
}

Missed this the first time around because in my deck, I defined font-weight: "normal" in my theme; missed it the second time around in unit tests for style refactoring, hit it while writing a "tricky" deck to find any misunderstandings I have about effects of theme, base props, specific props, style prop, default element formatting, and inherited formatting.

pedrottimark:
Please verify #106 that pacman progress remains a solid dot when you go forward or back a slide.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -230,7 +230,7 @@ These tags create a styled blockquote. Use them as follows:
 </BlockQuote>
 ```
 
-####\<CodePane />
+####\<CodePane /> (Base)
 
 This tag displays a styled, highlighted code preview. I prefer putting my code samples in external `.example` files and requiring them using `raw-loader` as shown in the demo. Here are the props:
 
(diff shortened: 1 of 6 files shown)


kenwheeler:
Thank you sir!

PR #110, opened 2015-11-22
Refactor

knowbody:
refactored some of the code, check if you like it @kenwheeler


src/appear.js
```diff
@@ -21,10 +21,10 @@ const Appear = React.createClass({
     };
   },
   componentWillMount() {
-    this.context.export || this.context.flux.stores.SlideStore.listen(this._storeChange);
+    return this.context.export || this.context.flux.stores.SlideStore.listen(this._storeChange);
```
kenwheeler: why retrun

knowbody: there must be some return in those functions, else eslint is complaining

kenwheeler:
Ok so everything looks pretty good, except for build. Build should wrap everything up into dist. HTML & images included. @knowbody

knowbody:
@kenwheeler forgot about that, will fix now. thanks

PR #113, opened 2015-11-23
Fix import statement for lodash/lang/isUndefined

MarkRedeman:
When running `npm start` I got an error that `lodash/lang/isundefined` could not be resolved (used in `code-pane.js` and `markdown.js`).
I'm not sure why it couldn't resolve it (the path seems to be right), but this PR fixes it.


kenwheeler:
Thanks dude

Issue #108, opened 2015-11-19
Refactor style: replace Base class with util functions


pedrottimark:
Proposal for replace inheritance part of #104

```
A. Goals:
1. Given a component and props, render elements with the same style as before.
2. Automagically correct minor inconsistencies and omissions in propTypes of component classes.
3. Test style util functions independently of Radium and React components.
   Also testing result of Radium mergeStyles would be bonus,
   but doing that independently of React components is beyond my current knowledge and skill.
4. Remove class inheritance as a potential barrier for users to add components (or install third-party components).
   Also remove as a barrier to React functional components if Radium would support them.
5. More easily add new props, correct problems with props, or refactor props (if needed).
```

See below for example for Link of current compared to refactored style API.
B. Main points of refactoring concept:

1. `styleCascade(styleArray, props, context, state)` returns same array of style objects as before.

2. References to props of `this.context.styles.components` become string or array

```javascript
.link becomes "link"
.codePane.pre becomes ["codePane", "pre"]
.heading[`h${this.props.size}`] becomes ["heading", (props) => `h${props.size}`]
```

3. `Base.getStyles` becomes `styleBase /* (props, context) */`

4. `this.props.style` becomes `styleProp /* (props) */`

5. `const styles = {flex: 1}` becomes `{flex: 1}`

6. `const styles = {width: this.props.width || "", ...}` becomes either

```javascript
{width: this.props.width || "", ...} // or recommended
styleImage /* (props) */ that is, a util or local function which returns a style object
```

7. Define util or local functions for fit case of Heading (or Text)

```javascript
[this.context.styles.components.text, this.getStyles(), styles.container] // and
[styles.text, this.props.style] // become

["text", styleBase, styleFitContainer /* ( , , state) */] // and
[styleFitContained /* (props, , state) */, styleProp]
```

8. `propTypesCascade` refers to `styleBase.propTypes`, `styleProp.propTypes`, and so on

9. Therefore, if you write a local `styleComponent` function, do one of the following:

```javascript
styleComponent.propTypes = { ... }; // the style props that it "consumes"

propTypesCascade(styleArray, { ... , children: PropTypes.node}); // if it has children, of course
propTypesCascade(styleArray, { ... }, {children: PropTypes.node}); // can separate style props from children if you think it is clearer
```

C. Example for Link
1. current component style API:

```javascript
import Base from "./base";

@Radium
class Link extends Base {
  render() {
    return (
[...]

PR #115, opened 2015-11-24
styleBase(this.props, this.context)

pedrottimark:
Follow up on #108 superseded by #121


pedrottimark:
If this is merged, I can submit a follow-up with tests/specs/base.spec.js

PR #124, opened 2015-12-04
PDF export section typo

MrMormon:


kenwheeler:
What an asburd typo. Thanks!

PR #126, opened 2015-12-06
Remove redux-logger import

fson:
createLogger was imported but not used. Fixes the error:
"Module not found: Error: Cannot resolve module 'redux-logger'"


kenwheeler:
Thanks dude!

PR #130, opened 2015-12-11
Use babel-core polyfill for Safari.

threeve:
Safari lacks es6 Object.assign. This uses the solution from
http://stackoverflow.com/a/32917654 to add the babel-core polyfill which
fixes support for Safari (and potentially other polyfills?).

Not entirely sure this is the right solution, but it does work for me. Tested by making the same changes to the webpack.config.js files in spectacle-boilerplate.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -7,6 +7,7 @@ module.exports = {
   devtool: 'source-map',
   entry: [
     'webpack-hot-middleware/client',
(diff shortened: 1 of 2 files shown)


knowbody:
good catch! @kenwheeler we want that :point_up:

kenwheeler:
@threeve Thanks so much!

PR #131, opened 2015-12-11
upgrade to babel 6

### LINKED ISSUE #118: Babel 6
@kenwheeler is it better to go ahead and get the change from 5 to 6 out of the way?

devDependencies is not a particular strength of mine, but I know somebody... @knowbody :)

knowbody:
Yay! Upgraded to Babel 6, there is one minor problem that Babel 6 does not support decorators yet (http://phabricator.babeljs.io/T2645), but there is a fix for now which is `babel-plugin-transform-decorators-legacy` (https://github.com/loganfsmyth/babel-plugin-transform-decorators-legacy)

if that's cool, closes #118


webpack.config.production.js
```diff
@@ -1,23 +1,23 @@
 /* eslint-disable */
```
ryan-roemer: How's about some eslint folks :wink:

webpack.config.production.js
```diff
@@ -1,23 +1,23 @@
...
 
-var path = require("path");
-var webpack = require("webpack");
+var path = require('path');
```
ryan-roemer: GLOBAL: `eslint-config-defaults` uses double quotes. We should revert _and_ get lint turned on for all JS files in this repo. (Can be separate file).

webpack.config.production.js
knowbody: this is me loving single quotes ;p I will update the PR

webpack.config.production.js
knowbody: @ryan-roemer is `/* eslint-disable no-var */` ok in this file?

ryan-roemer: @knowbody -- This should be linted as node code, not frontend React code. We should use https://github.com/walmartlabs/eslint-config-defaults/blob/master/configurations/walmart/es5-node.js for node files.

knowbody: oh okay but that should be in the separate PR, right? I'll just make sure that we use double quotes here.

ryan-roemer: As a helpful tip, take a look at how I linted a Spectacle presentation in https://github.com/ryan-roemer/surge2015/blob/master/package.json#L10-L12

Side note -- the lint may be too unwieldy / complicate this specific PR, so feel free to ticket out + revert the quotes thing and then have this PR just focus on the babel stuff.

Thanks!

ryan-roemer: > oh okay but that should be in the separate PR, right? I'll just make sure that we use double quotes here.

Haha -- simultaneous messages. **Sounds perfect** Thanks!

knowbody: Ha! I like your solution for your surge2015 presentation.

knowbody:
@kenwheeler is anything stopping this from merging?

kenwheeler:
Ship it!

PR #138, opened 2015-12-30
Remove references to react-router

will-hart:
`react-router` was removed in `1.0.0` but there are still references to it in `package.json` and the example slides. This pull request deletes references to `react-router`.


kenwheeler:
Thanks!

Issue #153, opened 2016-01-29
npm run build:dist fails


oliviertassinari:
Why is `transform-decorators-legacy` only enable in development? https://github.com/FormidableLabs/spectacle/blob/cdfd5c458deb9c6ea15f51df0fac91f52bac8fb6/.babelrc#L6

knowbody:
yeah, could be just:

```json
{
  "presets": ["es2015", "stage-0", "react"],
  "plugins": [
    ["transform-decorators-legacy"],
  ]
}
```

oliviertassinari:
I wanted to submit a PR another topic. Looks like it will be a two in one :).

PR #154, opened 2016-01-29
[babel] Improve the production build

oliviertassinari:
Fix https://github.com/FormidableLabs/spectacle/issues/153.
And add some plugins for improving the production build.

| Name | Build size impact | Description |
|:-----|:-----|:------|
| [transform-react-remove-prop-types](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types) *(Promoted)* | -10 kB | Remove `PropTypes`. | 
| [transform-react-constant-elements](https://babeljs.io/docs/plugins/transform-react-constant-elements/) | 0 kB | Save some memory allocation. |
| [transform-react-inline-elements](https://babeljs.io/docs/plugins/transform-react-inline-elements//) | +10 kB | Improve runtime performances. |
| [transform-runtime](https://babeljs.algolia.com/docs/usage/runtime//) | -80 kB | Remove duplicated babel helper functions. |


knowbody:
lgtm! :rocket:

kenwheeler:
Thanks so much!

package.json
```diff
@@ -10,7 +10,7 @@
     "build:dist": "NODE_ENV=production webpack --config webpack.config.production.js",
     "lint": "eslint --ext .js,.jsx src",
     "deploy": "npm run build:dist && surge -p .",
-    "start": "node server.js"
+    "start": "NODE_ENV=development node server.js"
```
ryan-roemer: Side Note -- this will break windows compatibility. The `build:dist` is probably OK for breaking, but might be nice to have this still work. Can we just leave _no_ `NODE_ENV` and have that be defaulted to be `development`?

(or, use `builder envs` ;) )

oliviertassinari: Oups, I have forgotten to remove this line from my commit. Feel free to fix it.

kenwheeler: Awesome catch @ryan-roemer , fixed here a32123c36ee9f47d6e7fbd61c3f95820d6ee43b8 

Does it make sense to build a Spectacle builder archetype for core too? Or just for the boilerplate?

ryan-roemer: @kenwheeler -- I think just add `builder` alone to get the extra's like `builder envs` and `builder concurrent` and not worry about an archetype since spectacle itself is kind of a "one-off".

The `spectacle-boilerplate` could definitely be replaced by an archetype like `builder-spectacle` now that `builder-init` has landed.

That's my gut intuition, but it _may_ be the case we could add `init/` templates and `init.js` to `spectacle` itself and just have it all together.

PR #162, opened 2016-03-12
Remove `#` from links

r24y:
I cloned the boilerplate and clicked the links in this README; did not have the intended effect. This commit updates the links so that the boilerplate works.


kenwheeler:
Thanks!

PR #163, opened 2016-03-20
update readme example to import Spectacle component rather than S

bencripps:
Small update to ReadMe to import `Spectacle` Component, rather than `S`. 

I thought this was better than

`import { S as Spectacle } from 'spectacle`


kenwheeler:
Thanks!

PR #166, opened 2016-04-01
Fix alignment typo in README

oksas:
The readme [states](https://github.com/FormidableLabs/spectacle#slide-base) that a slide's `align` property takes two values, both of which can either by `flex-start` or `center` or `flex-end`, but then states that the default is `align="center middle"`, which should actually be `align="center center"` I think, so that change has been made. Also changed `it's` to `its` on line 241 (the same line).

Spectacle is awesome, by the way, thanks for making this :+1:


kenwheeler:
Thank you!

Issue #57, opened 2015-08-22
Allow SlideSets to be defined


bebraw:
Currently you are forced to maintain all your `Slides` in a single `Deck`. It would be useful if it was possible to defined `SlideSets` like this:

``` html
<Deck ...>
  <SlideSet ...>
    <Slide ...>...</Slide>
    ...
  </SlideSet>
  ...
</Deck>
```

After this you could push `SlideSets` to outside of the main file easily. As the current system expects `Slides` to belong directly below `Deck` this is hard/impossible to achieve (yup, tried _many_ ways).

MarkRedeman:
For anyone else wanting to separate their slides from their decks, this is how I've done it (it is similar to how #32 has implemented extension of slides):

First I've extended the `Slide` component so that it will render the ouput of a `contents()` method.
```
import {Slide as SpectacleSlide} from "../src/spectacle";

/**
 * I'm using this class so that in deck.js I can use small independent slide components
 * Since most of my slides will use a default transition I also provide some default props
 */
export default class Slide extends SpectacleSlide {
    render() {
      this.props.children = this.contents();
      return super.render();
    }
}

Slide.defaultProps = {
  ...SpectacleSlide.defaultProps,
  transition: ["slide", "fade"]
};
```

Then if I want to create a new slide I can do this:
```
export class Title extends Slide {
  contents() {
    return (
      <div>
        <Heading size={1} caps textColor="primary" fit>
          My Title
        </Heading>
      </div>
    );
  }
}
Title.defaultProps = {
  ...Slide.defaultProps,
  bgImage: images.dna.replace("/", ""),
  bgDarken: 0.5
};

export class AboutMe extends Slide {
  contents() {
    return (
      <div>
        <Heading size={1} caps textColor="black" fit>
          @MarkRedeman
        </Heading>
      </div>
    );
  }
}
```

Next in `deck.js` I import all slides and render them one at a time:
```
import * as Introduction from "./introduction";

// other stuff
render() {
    return (
        <Deck transition={["slide", "fade"]} transitionDuration={500} progress="pacman">
                <Introduction.Title notes="Title notes" />
                <Introduction.AboutMe notes="About me notes." />
        </Deck>
    );
}
```

firstdoit:
This is a nice hack, but `SlideSet`s would definitely be simpler.

kenwheeler:
[...]

PR #168, opened 2016-04-28
Enable prev and next slide with up and down keys

patrick91:
Shall I make this configurable somehow? Or is it fine as it is now?

### DIFF
--- a/src/components/deck.js
+++ b/src/components/deck.js
@@ -86,13 +86,18 @@ export default class Deck extends Component {
   _handleEvent(e) {
     const event = window.event ? window.event : e;
 
-    if (event.keyCode === 37 || event.keyCode === 33 || (event.keyCode === 32 && event.shiftKey)) {
+    const prevKeys = [37, 38, 33];
+    const nextKeys = [39, 40, 34];
+
+    const { keyCode } = event;
+
+    if (prevKeys.indexOf(keyCode) >= 0 || (keyCode === 32 && event.shiftKey)) {
       this._prevSlide();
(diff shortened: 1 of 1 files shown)


ryan-roemer:
@kenwheeler -- Where did we end up with the "2d matrix" idea of slides like reveal does it? (If we're still toying with that may want up/down to be saved for just within a column)

patrick91:
@ryan-roemer I'd make it at least configurable if we have something like reveal :)

My [remote](https://www.amazon.co.uk/Doosl-DSIT001B-IT11-DSIT001/dp/B013W64932?ie=UTF8&psc=1&redirect=true&ref_=oh_aui_detailpage_o01_s00) has only up/down buttons :\

kenwheeler:
I definitely want the 2d matrix at some point. That said, it would be cool to do an atom style key map config to support a variety of remotes.

kenwheeler:
Electron style*

patrick91:
@kenwheeler can you point me at some examples? I wasn't able to find anything.

bebraw:
Maybe it would make sense to expose the keys through a `prop` while using the current defaults. You could also consider mapping strings to ASCII to make it easy to use.

👍  to reveal.js style slide sets. That would be awesome but also a separate PR.

halhenke:
Glad i checked this PR - was thinking of taking a crack at the Reveal style grid as in #57 and was wondering how people would feel about the change in key bindings (in Reveal the left and right arrows switch between "stacks of slides" while up and down goes between slides/fragments). 

Making it configurable should make everyone happy.

kenwheeler:
Closing in favor of a feature discussion regarding keymapping

PR #196, opened 2016-09-01
Allow arrow keys up and down to control transition

welcoMattic:
It works better for pointing device that support only up and down keys.

### DIFF
--- a/src/components/deck.js
+++ b/src/components/deck.js
@@ -88,9 +88,9 @@ export default class Deck extends Component {
   _handleEvent(e) {
     const event = window.event ? window.event : e;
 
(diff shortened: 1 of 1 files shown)


ryan-roemer:
It is probably best to leave up / down for the 2-d grid / "slide sets" feature if we still want to pursue that per https://github.com/FormidableLabs/spectacle/issues/57 for competitiveness with Reveal.js and other presentation frameworks that support 2-d grids...

welcoMattic:
I didn't know that 2-d grid / slide sets feature is planned. It's looks ok to me to close my PR ;) No regrets

PR #199, opened 2016-09-12
reconfiguring the example file so that it builds and is easier to export

ebrillhart:
This reconfigures the file structure that holds the live Spectacle example and updates the build script to include it. This will ease the transfer of the example from its current location at stack.formidable.com to formidable.com/open-source/spectacle/example. Hopefully this works/the example was not previously excluded from the build for an important reason. 

I think an npm publish will be required once this is merged, and I'm happy to go through that process if pre-approved to do so!

cc/ @ryan-roemer @bmathews @kenwheeler


ryan-roemer:
I think this needs updates to the ignore files as well. Gitignore and npmignore to be consistent with the existing source and lib directories

ebrillhart:
@ryan-roemer like so?

.gitignore
```diff
@@ -1,4 +1,5 @@
 node_modules
 dist
 lib
+example/lib
```
ebrillhart: @ryan-roemer like so?

bmathews: yeah, not needed, since the prior `lib` will cover it

.npmignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 src
+example/src
```
bmathews: npmignore follows same rules as gitignore. prior `src` will cover `example/src`, so this is not needed

ebrillhart:
@ryan-roemer @bmathews good to go once I revert the gitignore and npmignore back?

.npmignore
```diff
@@ -1,3 +1,3 @@
 node_modules
 src
-*.log
+*.log
```
ryan-roemer: Let's get that newline back too 😉

ryan-roemer:
@ebrillhart -- One nit, after which LGTM. No re-review needed.

PR #201, opened 2016-09-16
Don't set documentElement fontSize when slide viewerScaleMode is true

bmathews:
Global mutation of documentElement style was breaking the site containing the spectacle presentation. If I'm understanding the slide's `viewerScaleMode` prop correctly, then when it's true, setting the documentElement's fontSize won't make a difference anyway.


kenwheeler:
thanks bmats @bmathews

PR #208, opened 2016-10-02
Add support for ordered list in `List`component

jonasws:
This PR adds ordered list functionality to the `List` component, in order to enable functionality as described [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol). Supported props in ordered lists are:

* reversed
* start (index to start numbering)
* type (lettered, numbered, Roman numerals...)

## Usage
```jsx
// Capital letter list starting from 2
<List ordered start={2} type="A">
  <ListItem></ListItem>
</List>
```

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -354,8 +354,26 @@ The link tag is used to render `<a>` tags. It accepts an `href` prop:
 <a name="list--listitem-base"></a>
 #### List & ListItem (Base)
 
+|Name|PropType|Description|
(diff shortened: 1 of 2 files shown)


waywaaard:
Good addition, especially the support for the start attribute.

jonasws:
@kenwheeler Care to take a look?

kenwheeler:
Thanks so much!

PR #220, opened 2016-10-21
Accept arrays of `Slide` in `Deck`

fand:
Currently we cannot pass arrays of `<Slide>`s to `<Deck>`, such as

```js
<Spectacle theme={theme}>
  <Deck transition={["zoom", "slide"]} transitionDuration={200} progress="bar">
    {mds.map(md => <Markdown source={md}/>)}
  </Deck>
</Spectacle>
```

It would be great if we could!

## Problem
`<Deck>` treats `this.props.children` directly.
`this.props.children` may include arrays as the element like:
`[<Slide/>, [<Slide/>, <Slide/>], <Slide/>]`

So `this.props.children.length` returns `3` while it actually has `4` slides.

## Solution

However, in some methods `Deck` treats children with `React.Children`.

```js
import { Children } from 'react';

assert.deepEqual(this.props.children, [<Slide/>, [<Slide/>, <Slide/>], <Slide/>]);
this.props.children.length; // 3;

const children = Children.toArray(this.props.children); // [<Slide/>, <Slide/>, <Slide/>, <Slide/>]
children.length // 4
```

I fixed `Deck` to treat children via `React.Children` so that we can pass arrays of `<Slide>`s.

### DIFF
--- a/src/components/deck.js
+++ b/src/components/deck.js
@@ -163,13 +163,14 @@ export default class Deck extends Component {
     this.setState({
       lastSlide: slide
     });
+    const children = Children.toArray(this.props.children);
     if (this._checkFragments(this.props.route.slide, true) || this.props.route.params.indexOf("overview") !== -1) {
-      if (slide < this.props.children.length - 1) {
+      if (slide < children.length - 1) {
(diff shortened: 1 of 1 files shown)


kenwheeler:
Thanks!

PR #235, opened 2016-10-25
Adds first markdown-it Markdown Component

ebertmi:
This is work in progress but features a fully working Markdown Component that is based on markdown-it.

Can basically do everything that the previous component did:
- Extensible, plug-in your own `onIterate` function to override the components
- Allows to use `markdown-it` plugins
- Supports inline html (but does not support Markdown inside inline html)
- Built-in support for `markdown-it-decorate` and supports html inline styles
- The default `onIterate` does also supports math tags, if configured

image


kenwheeler:
This looks pretty great, @carlospaelinck can you check this out?

waywaaard:
I rebased on the latest changes and added `Object.assign`.

waywaaard:
I tried to fix most of the errors. Using the variable name `props` caused a lot of errors... I did not fix *too many parameters*, the * too many statements*  and the *complexity* stuff that is caused by the recursiv tree iterating.

waywaaard:
I just discovered a bug using the Markdown-Component inside an `<Appear>`. It seems that I am somewhere missing to pass the style or to receive the style from the parent, hence the appear is not working.

Simple example:
```jsx
          <Slide transition={["spin", "zoom"]} bgColor="tertiary">
            <Heading caps fit size={1} textColor="primary">
              Inline Markdown
            </Heading>
            <Markdown>
              {`
![Markdown Logo](${images.markdown})

You can write inline images, [Markdown Links](http://commonmark.org), paragraph text and most other markdown syntax
* Lists too!
* With ~~strikethrough~~ and _italic_
* And let's not forget **bold**
* And basic inline HTML support: <kbd style="vertical-align: middle; display: inline-block; padding: 3px 5px; border: solid 1px #ccc;, border-radius: 3px; border-bottom-color: #bbb; box-shadow: inset 0 -1px 0 #bbb">Ctrl</kbd> + <kbd style="vertical-align: middle; display: inline-block; padding: 3px 5px; border: solid 1px #ccc;, border-radius: 3px; border-bottom-color: #bbb; box-shadow: inset 0 -1px 0 #bbb">Yeah</kbd>
              `}
            </Markdown>
            <Appear>
              <Markdown>{"And some more Markdown in an `<Appear>` tag."}</Markdown>
            </Appear>
          </Slide>
```

carlospaelinck:
I've noticed that as well before and after the conversion to `VictoryAnimation`. Wrapping `<Markdown />` in a `<div>` appears to do the trick. Not sure why yet.

waywaaard:
But wrapping the markdown component would prevent the styles to be passed down, right?

PR #237, opened 2016-10-26
Switch to react-history module.

### LINKED ISSUE #228: Switch to react-history
We should move away from the `history` module, and instead use https://github.com/mjackson/react-history

carlospaelinck:
* Use the React History hash history component instead of history
* Removes an additional listener in the controller

Fixes #228 

// cc: @kenwheeler


waywaaard:
Does this also fix the problem, that spectacle does always show the first slide even if you link to a specific slide with `#6` or so?

I've added support in the controller for my local branch because this was quite annoying.

carlospaelinck:
You're right. I've added support to enter into Spectacle with any valid slide index with checking for too big or negative numbers.

Issue #226, opened 2016-10-25
Switch to react-motion


kenwheeler:
We should switch from react-tween-state to react-motion.

cc @carlospaelinck

PR #241, opened 2016-11-01
Update animations to use VictoryAnimation.

carlospaelinck:
* Create decorator for Transitionable slides
* Create decorator for Slide render
* Convert Slide and Appear to ES6 class

Addresses #226 

// cc: @kenwheeler


kenwheeler:
well done

PR #245, opened 2016-11-08
Use CommonJS Modules Plugin for Build Lib.

carlospaelinck:
* Ensure common JS modules are used for building the lib

// cc: @kenwheeler


package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
```
ryan-roemer: NIT: In this pr or another, let's remove:

```js
    "eslint-config-defaults": "^9.0.0",
```

carlospaelinck: 👍

Issue #233, opened 2016-10-25
Investigate integration with typography.js


kenwheeler:
https://kyleamathews.github.io/typography.js/

This could help users with font/style choices

PR #248, opened 2016-11-15
Add Typeface tag and Google Font support to themes.

carlospaelinck:
* Themes can now load a system font or Google Font
* Google Font style tags will be automatically created and appended to DOM
* `<Typeface />` tag can be used for specific font styling of text
* Added and documented changes and examples to README

Addresses #233 

// cc: @kenwheeler


carlospaelinck:
LGTM

PR #255, opened 2016-12-10
Initial Support For Remote Control/Viewing

mattvoss:
This adds initial support for remote control of presentations. The example implementation uses websockets however any socket type could be used theoretically. 

It works by transmitting the state on a change action to any connected node. All nodes can control the presentation. However the props specifying presenter mode and overview mode are stripped from the state object prior to transmission. This prevents the node where those modes are invoked from propagating those modes to the other connected nodes.

There is a new action added "REMOTE_STATE" to support this functionality. 

In addition the last action called is stored in state. I think this should be reviewed and may not be necessary if a more correct "Redux" way could be found.

The server has been altered to support websockets and contains some new functionality to support remote control/viewing. One thing to note is that the latest state is saved on the server to allow newly connected nodes to sync to the most current state of the presentation. Right now to completely reset state one of two things must be done. Either restart the server or go back to the very beginning of a presentation.

### DIFF
--- a/server.js
+++ b/server.js
@@ -45,7 +45,6 @@ wss.on('connection', function connection(ws) {
   // or ws.upgradeReq.headers.cookie (see http://stackoverflow.com/a/16395220/151312)
   ws.on('message', function incoming(message) {
     const payload = JSON.parse(message);
-    // console.log('received: ', util.inspect(message, false, null));
     state = payload;
     wss.clients.forEach(function (client) {
       if (client !== ws) client.send(message);
@@ -59,11 +58,6 @@ wss.on('connection', function connection(ws) {
   ws.send(
     JSON.stringify(state)
   );
-  /**
-  state.forEach(function (op) {
-    ws.send(op);
-  });
-  **/
 });
 
 server.on('request', app);
--- a/server.js
+++ b/server.js
@@ -1,12 +1,32 @@
 /* eslint-disable */
-
+var server = require('http').createServer();
+var WebSocketServer = require('uws').Server;
+var wss = new WebSocketServer({ server: server });
+var url = require('url');
 var path = require("path");
 var express = require("express");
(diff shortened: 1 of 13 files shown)


kenwheeler:
This is great!

server.js
```diff
@@ -19,11 +39,42 @@ app.get("*", function(req, res) {
...
-  console.log("Listening at http://localhost:3000");
+  ws.on('close', function close() {
+    console.log('socket disconnected');
+  });
+
+  ws.send(
+    JSON.stringify(state)
+  );
+  /**
+  state.forEach(function (op) {
```
MarkRedeman: Did you overlook this comment?

kenwheeler: I didn't fully review yet. We are starting work on a mobile presenter remote app, and I want the person tasked to have a look too. good catch.

MarkRedeman:
Neat!

I'm assuming this won't work when you deploy the slides to surge with `npm run deploy`? (Maybe add this remark to the documentation?)

Would it be difficult to add a constraint so that only the presenter can change slides? Perhaps we could add an option so that participants can either "follow" the presenter (and use the state from the websocket), or manually look through the slides without changing the slides for other participants.

server.js
mattvoss: Actually that is cruft that I should have removed before making the commit. My fault there. Anyways this pull request will need to be refactored in light of the pending removal of the spectacle tag.

jpwesselink:
@MarkRedeman Using a secret would be an easy way of to prevent non-presenting users from running the show. Have you observed how this was solved in revealjs?

kale-stew:
Closing this PR as it's now 2 years old and we're working through a rewrite to make this easier in a newer version of Spectacle. Thank you for all of your work on this, @mattvoss, you've helped lay the foundation for how this feature is going to work in the near future. 💯

PR #257, opened 2016-12-15
Add server side rendering support

Creeplays:


kenwheeler:
We allow custom history to be passed in now, closing in favor of that approach

Issue #214, opened 2016-10-12
Speaker notes are too small to be useful


0xR:
I like to have big bullets points on my screen to ensure I don't forget things. Because the notes are currently so small this causes me pause presenting trying to try to read from the screen. This ruins the presentation.

Now I'm not using the notes and have to make cue cards.

0xR:
Thanks for addressing this!

PR #262, opened 2016-12-29
Reorganize presenter screen to make room for notes

rawrmonstar:
Addresses #214 

This reorganizes the presenter screen to make more room for slide notes:

image


kenwheeler:
👍

PR #285, opened 2017-02-14
Changes Overview to show Slides and SlideSets as a 2D grid

zachhale:
This updates the overview view to show `SlideSet`s as vertical stacks and allow an up/down/left/right keyboard navigation to navigate slides in a grid (see below).

* on normal presentation and presenter views, keyboard navigation is unaffected. Left/right go forward-next in 1D series
* on overview, left/right/up/down navigates a 2D grid of slides. If you use space/shift+space you'll advance as you would in the presenter view. If you use arrow keys, you will navigate the grid.

Overview if most slides are simple siblings:
image

Overview with `SlideSet` organization:
image

Related to #57

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
   "license": "MIT",
   "dependencies": {
     "component-playground": "^1.3.1",
+    "dom-scroll-into-view": "^1.2.1",
     "lodash": "^4.16.4",
     "markdown-to-react-components": "^0.2.3",
     "mdast": "^2.1.0",
--- a/src/components/__snapshots__/manager.test.js.snap
+++ b/src/components/__snapshots__/manager.test.js.snap
@@ -412,6 +412,9 @@ exports[`<Manager /> should render the overview configuration when specified. 1`
           data-radium={true}
           style={
             Object {
+              "display": "flex",
+              "flexDirection": "row",
+              "flexWrap": "nowrap",
               "height": "100%",
               "overflow": "scroll",
               "width": "100%",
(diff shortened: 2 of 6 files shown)


kenwheeler:
@stefvhuynh can you review and merge this too if all good?

stefvhuynh:
@kenwheeler, there are some pretty gnarly merge conflicts in this branch. is @zachhale around to help sort it out?

jonasws:
@stefvhuynh @kenwheeler Is this getting any love lately? I'd be happy to help out if the original submitter is busy or "dosn't need" this feature as badly anymore.

ryan-roemer:
@jonasws -- Thanks so much for throwing your hat in the ring! If you've got some bandwidth, we'd love to get your help to kick this one into shippable shape (either via a from scratch PR or extending this one).

I'm happy to review / kick the tires anywhere along the way (_disclaimer_ I absolutely loved Reveal.js having this feature from back in the day, so I excited to see it in spectacle) and I can rope in @kenwheeler , @stefvhuynh , and co for nitty-gritty technical feedback when needed.

jonasws:
Do we also want to be able to use up/down to navigate slide sets while in presenter mode, similar to how it is done with reveal.js? As far as I know it would not conflict with any existing behavior.

ryan-roemer:
@jonasws -- Yep! The existing PR has navigation support that mostly brings in the reveal conventions.

I'm not sure if we're exactly following the Reveal.js shortcuts (https://github.com/hakimel/reveal.js/wiki/Keyboard-Shortcuts), so if you'd like to propose a different set of keys that (1) are sensible for legacy 1-d sequence, and (2) enable good 2-d slide navigation and overview slide jumping, go for it!

jonasws:
@ryan-roemer As the `SlideSet` components at the time is actually not documented, is there any behavior related to it that we actually need to keep? In terms of backwards compatibility with linear navigation etc? I figured it may be a smooth solution to have `SlideSet` work as a "vertical grouping" of Slides.

If we want this behavior to be optional, I'm sure one can do this with a prop ala `verticalGroup`, which would then default to `false`.

jonasws:
@ryan-roemer I've tried to get this working, but have failed to achieve some of the nitty-gritty details. I've heard talk about Spectacle eventually getting a "custom keymap" feature, and I think one should revisit this issue when that is in place. At the moment there is not a good way to make this backwards compatible, and make it play along nicely with third-party components such as the Terminal component. I think it would be wise to start this from scratch, as the progress of the codebase has overtaken me in trying to get this all working. You may want to close this PR to indicate this is not something being currently worked on, and potentially create an issue with a "feature request" label. That way the community can all join in and discuss how to best implement such a major feature.

PR #289, opened 2017-02-27
Allow ability to override duration of transition in `Appear` component

jlongster:
It would be nice customize the duration of the `Appear` component so I can make weird custom animations!


src/components/appear.js
```diff
@@ -36,11 +36,12 @@ class Appear extends Component {
   render() {
     const child = React.Children.only(this.props.children);
     const endValue = this.state.active ? 1 : 0;
+    const transitionDuration = this.props.transitionDuration || 300;
```
waywaaard: I think the proper way would be to use reacts defaultProps to set the transitionDuration, e.g.

```javascript 
Appear.defaultProps = { 
    transitionDuration : 300
};
```

jlongster:
Responded to review!

src/components/appear.js
```diff
@@ -57,6 +58,10 @@ class Appear extends Component {
   }
 }
 
+Appear.defaultProps = {
+  transitionDuration: 300
+};
+
 Appear.propTypes = {
```
waywaaard: `transitionDuration` needs to be added to the propTypes. Travis error:

```
'transitionDuration' is missing in props validation  react/prop-types
```

jlongster:
Sorry for the churn! I'll see if the tests pass this time myself.

jlongster:
Yay they passed! Thanks for the quick review. Let me know if there is anything else.

carlospaelinck:
This looks good to me 
// cc: @kenwheeler

PR #290, opened 2017-03-09
Allow using S without a type.

Schoonology:
Sorry for submitting a PR without any prior discussion; this is an issue that hits me every time I come at Spectacle again after a month of not writing talks.

Previously, S required a type, even when used only for Base properties like `textColor`. Setting text to `""` or `"normal"` was a workaround; this change removes the requirement altogether.

### DIFF
--- a/src/components/__snapshots__/s.test.js.snap
+++ b/src/components/__snapshots__/s.test.js.snap
@@ -1,3 +1,18 @@
+exports[`<S /> should allow custom styling 1`] = `
+<S
+  textColor="magenta">
+  <span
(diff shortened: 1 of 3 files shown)


Schoonology:
On an unrelated note, what is the best way to `npm link` Spectacle into a Spectacle Boilerplate presentation for local development?

Schoonology:
I'm not sure what the failures are. There are local test failures, but they're not different than the results on `master`.

kenwheeler:
This should be resolved upstream with the emotion conversion

PR #302, opened 2017-04-20
Fix markdown.js codepane bug.

sjstyle:
This pull request fixes CodePane render bug when using markdown syntax.

### Sample presentation code 

Note : I attached `Presentation` class without import clause. You can paste and run this code at example/index.js.

``` javascript
export default class Presentation extends React.Component {
  render() {
    return (
      <Deck transition={["zoom", "slide"]} theme={theme} transitionDuration={500}>
        <Slide transition={["zoom"]} bgColor="white">
          <Markdown>
              {`
  ## CodePane
---

  \`\`\`
  console.log("Hello world!");
  \`\`\`
            `}
          </Markdown>
        </Slide>
      </Deck>
    );
  }
}
```

At current master branch )

image

After fixes branch )

image

This happens because CodePane needs source with `props.source` or `children`, but marked.js (markdown renderer) returns `code` property to implement them.

I fix additional function by changing property `code` to `source`, so we don't modify `CodePane` nor `marked.js` source.

### DIFF
--- a/src/components/markdown.js
+++ b/src/components/markdown.js
@@ -28,10 +28,13 @@ const _S = (type) => {
 const _CombineBlockQuote = ({ children }) => <BlockQuote><Quote>{children}</Quote></BlockQuote>;
 _CombineBlockQuote.propTypes = { children: PropTypes.node };
 
+const _CodePane = ({ language, code }) => <CodePane lang={language} source={code}/>;
+_CodePane.propTypes = { code: PropTypes.string, language: PropTypes.string };
+
 markdownToReact.configure({
   a: Link,
   blockquote: _CombineBlockQuote,
-  code: CodePane,
+  code: _CodePane,
   del: _S("strikethrough"),
   em: _S("italic"),
   h1: _Heading(1),


mAAdhaTTah:
This should fix #329.

kenwheeler:
Can you rebase this to use the latest markdown style on master?

mAAdhaTTah:
@kenwheeler @sjstyle I integrated the above changes + changes to `Code` for the Markdown component but I'm getting a new error in my code. See here: https://github.com/mAAdhaTTah/spectacle/commit/a71765c3b3b217f9dba4b58d1d04e2ef6b8a50a2

sjstyle:
I noticed that this issue was solved by other commits. So I will close this pull request. 

I think #329 issue is about `<MarkdownSlides/>` tag's children attribute, so I found it is more complicated because marksy renders it's children markup tag. (`<Markdown/>` tag in spectacle's example)

sjstyle:
@mAAdhaTTah 
I opened another pull request #338, which relates 329 issue.

PR #309, opened 2017-04-28
Add Node v7 to CI config

exogen:
Checking if Node 7 has better luck on Travis than Node 5.


exogen:
Looks like #308 fixes Node v5. I'm going to change this to just add Node v7 instead of replacing v5.

kenwheeler:
LGTM

PR #313, opened 2017-05-09
[Proposal] Replace component-playground with react-live

philpl:
This should do a couple of things:

- Shrink the bundle (No Codemirror, or Babel Standalone)
- Use Prism for both: The code editor and the `CodePane`
- Snappier live editing

This would also open up the possibility of improving the `CodePane` based on the Editor component in React Live.

cc @kenwheeler


philpl:
Talked to @carlospaelinck and this PR will be superseded by an up-to-date one

PR #325, opened 2017-06-02
Notes component

### LINKED ISSUE #322: Notes Component
Are you open to the idea of a `Notes` component that would allow adding notes within the slide instead of as a prop. This would allow for leveraging JSX for the notes. 

I would expect that the `Slide` would use (something like) `React.Children.toArray().filter(c=> c!=Notes)`

codepunkt:
I found it hard to have elaborate notes with lots of html as an attribute of `Slide`, so i added a `Notes` component.

I had to add `cross-env` to make it build on windows.
Also, i just found out that i'm not alone and this probably fixes #322.

Any input is welcome. Thanks for the lib! 👍


kenwheeler:
Well done

nuragic:
@kenwheeler could this one be released? `<Notes>` is not available in the last `v2.1.0` 😕 ... Thanks!

PR #330, opened 2017-06-13
Fix direction of slide transition

philipphenkel:
Defect since introduction of SlideSet in 49b191723591f8aa4f18004a3268256298cde720 in January.

The wrong transition direction is especially visible on forward "slide" transitions.


delvedor:
Hi! When this fix will be released? :)

PR #332, opened 2017-06-18
Refined documentation for CodePane language support.

martinlechner1:


travi:
might be worth also including the link to the list of supported languages: http://prismjs.com/#languages-list

PR #339, opened 2017-07-02
Fix transition direction of slides with string based hashes

philipphenkel:
Use the manager’s slide reference to properly look up the route’s slide index.

The "slide" transition animation is broken when moving backward in slides with string based id hashes. The wrong transition direction gets visible if two slides with string hashes are in a row, transition "slide" is used and you are navigating backward.


philipphenkel:
This PR also fixes the broken snapshot of the markdown slide.

PR #353, opened 2017-08-28
Pass child css through to element wrapped in Appear tag

knittingcodemonkey:
Fixed a bug where css was lost when applied to an element contained inside an <Appear /> tag.
Also added an example to show how one can use this new ability to layer fragments on top of each other with css.


kenwheeler:
Thanks!

PR #360, opened 2017-09-22
Step counter feature

Grimones:
Hi

This PR implements step counter for `<Appear />` in `<Slide />`
To use simply add `getAppearStep` prop to `<Slide />` which is a callback and returns steps.

`<Slide ... getAppearStep={ steps => /* do something */ } /> `

Also added a slide to example.

Based on this branch https://github.com/pedrottimark/spectacle/tree/step-visitor

Use case can be like mine. I'm writing a presentation about ThreeJS and i have a canvas with renderer on slide background. On each step of text appearing i want to do something in ThreeJS.
I have a component with ThreeJS scene
`<CubeDemo step={this.state.cubeDemoStep} />`
And on each step i do something different in scene
```
animate() {
    const delta = this.clock.getDelta();
    if (this.step === 1) {
      this.mesh.rotation.x += delta;
    }
    else if (this.step === 2) {
      this.mesh.rotation.z += delta;
    }
    else if (this.step === 3) {
      this.mesh.rotation.x += delta;
      this.mesh.rotation.z += delta;
    }
}

### DIFF
--- a/example/src/index.js
+++ b/example/src/index.js
@@ -1,38 +1,38 @@
-import React from "react";
+import React from 'react';
 
 import {
   Appear, BlockQuote, Cite, CodePane, ComponentPlayground, Deck, Fill,
   Heading, Image, Layout, Link, ListItem, List, Markdown, MarkdownSlides, Quote, Slide, SlideSet,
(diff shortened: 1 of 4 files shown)


kenwheeler:
👍

PR #361, opened 2017-09-26
Accept strings for contentWidth and contentHeight

karlhorky:
To allow units such as `85vw`.

---

Hm, did I miss a reason why this won't work? Like causing problems with the presenter mode or different contexts where the slides will be displayed?


karlhorky:
Yeah, looks like I missed some things. It's expecting number values here:

https://github.com/karlhorky/spectacle/blob/content-dimensions-strings/src/components/manager.js#L214
https://github.com/karlhorky/spectacle/blob/content-dimensions-strings/src/components/manager.js#L532-L533

I guess there's probably not a desire to refactor these for this feature...

kenwheeler:
Yeah that will be tough to work around. Closing for now

karlhorky:
No problem, if this is a desired feature then this can be reopened or another PR started.

maxcan:
Currently the only place contentWidth and contentHeight are used for arithmetic in the code is here: https://github.com/FormidableLabs/spectacle/blob/f76d3bc246a5e4930bbbfc7db3e79d0afc826820/src/components/manager.js#L802

Can we replace that with a call to css `calc(..)`. have a branch but need to work on getting the tests to pass.

kale-stew:
@maxcan I'm happy to look at your branch with you. I know `contentWidth` is further reaching that it initially appears, I've had to modify this prop in the past, so I might be able to point you in the right direction?

Issue #365, opened 2017-10-05
Slides with large images can cover notes in presenter mode


adamgruber:
I noticed that when you have a large image on a slide the preview in presenter mode can become really wide, covering any notes you may have.

![screen shot 2017-10-05 at 9 58 39 am](https://user-images.githubusercontent.com/2378900/31231501-111addf4-a9b5-11e7-830b-485b6b7885f4.png)
![screen shot 2017-10-05 at 9 58 42 am](https://user-images.githubusercontent.com/2378900/31231499-10e77036-a9b5-11e7-8b13-348a2e1da605.png)

Demo repo here:
https://github.com/adamgruber/spectacle-presenter-issue

PR #367, opened 2017-10-09
Set max width of presenter current slide view. 

mihshhehl:
This prevents the current slide from overlapping with slide notes. Let me know if this is on the right track for a fix on #365


mihshhehl:
Looks like this is fixed with #357

PR #370, opened 2017-10-13
Added .example loader to webpack config files

zurfyx:
## Why?

>  I prefer putting my code samples in external .example files and requiring them using raw-loader as shown in the demo - https://github.com/FormidableLabs/spectacle#codepane-base

While in the example `source={require('raw-loader!../assets/deck.example')}` works just fine, we get a linter error when working with create-react-app by default. It can be hidden with /* eslint-disable */ ... /* eslint-enable */ but I don't think that's the best solution.

```
Line 24:  Unexpected '!' in 'raw-loader!./assets/hello.example'. Do not use import syntax to configure webpack loaders  import/no-webpack-loader-syntax
```

Since you are recommending `.example` file as source code example files, I thought it would be interesting to have `.example` files to be loaded with the `raw-loader` by deault.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -61,6 +61,10 @@ module.exports = {
         loader: 'url-loader?mimetype=image/jpg',
         include: path.join(__dirname, 'example/assets'),
       },
+      {
+        test: /\.example$/,
+        loader: 'raw-loader'
+      }
     ],
   },
 };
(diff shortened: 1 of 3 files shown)


kenwheeler:
This is probably better configured in spectacle-scripts or spectacle-boilerplate, as this repo isnt a base for authoring presentations. I'll get it updated in those, thanks!

PR #377, opened 2017-10-19
Start transition to Emotion. Fix Prism.

carlospaelinck:
- [x] Convert styled components to Emotion.
- [x] Update snapshots
- [x] Fix Prism to version `1.8.1` since the latest `1.8.3` is broken on the CDN.


mAAdhaTTah:
What's the issue w/ 1.8.3?

carlospaelinck:
@mAAdhaTTah The file is incomplete. We get an unexpected end of line.

PR #379, opened 2017-10-24
Radium to emotion

carlospaelinck:
- [x] Convert `@Radium` decorators for components to `react-emotion` components
- [x] Refactor `Slide` transitions and simplify logic
- [x] Update tests


ebertmi:
What is the reasoning behind switching from radium to emotion?

carlospaelinck:
@ebertmi This is the first phase in rewriting the themeing story and also a general refactor. We want to move to a modern css-in-js styling solution. Then look at where things are in how to make theming more approachable.

src/components/text.js
```diff
@@ -1,9 +1,40 @@
...
+    padding: '0',
+    lineHeight,
+    transform: `scale(${scale})`,
+    transformOrigin: 'center top'
+  },
+  styles.typeface,
+  styles.user
+]);
+
+const UnfitText = styled.p(({ lineHeight, styles }) => [
```
kenwheeler: THIS TEXT IS UNFIT

kenwheeler:
OK hold up, got some prollems. I'll send you videos

PR #383, opened 2017-10-28
Update README.md

kitze:
BREAKING CHANGE: It's no longer allowed to omit the '-loader' suffix when using loaders.
                 You need to specify 'raw-loader' instead of 'raw',
                 see https://webpack.js.org/guides/migrating/#automatic-loader-module-name-extension-removed


kenwheeler:
Thanks kitz

PR #387, opened 2017-10-30
Replace component-playground with react-live

philpl:
- Update README
- Replace playground.css with (emotion) styled react-live components
- Replace codemirror themes with DuoTone prism themes modified for
react-live
- Fix up playground default code
- Fix up playground tests and snapshots
- Remove codemirror and playground CSS

As a next step we could get rid of the CDN prism bundle and instead use a disabled React Live Editor. This might need some modifications and we'd have to pull in more parts of Prism, but react-live's usage of prism would allow this: https://github.com/FormidableLabs/react-live/blob/master/src/utils/prism.js


kenwheeler:
Dude. Amazing.

PR #389, opened 2017-10-31
Bundle Prism and use react-live's Editor instead

### LINKED ISSUE #388: Bundle prism and use react-live's Editor inside the CodePane component
Depends on #387 

If we're going to use react-live we can refactor the `CodePane` component to use a bundled version of Prism instead. This avoids duplication of any react-live code and should make it unnecessary to include Prism from a CDN

cc @carlospaelinck

philpl:
Fix #388 

- Remove all references to prism's CSS and CDN bundle from the readme
- Add prismjs package
- Use react-live's `1.8.0-1` i.e. `spectacletest` release which includes a `language` prop and externalised prism build
- Add `src/utils/prism-import` file (**This includes all syntaxes that we'll support out of the box**)
- Mirror ComponentPlayground theming and setup
- Rename theme properties

**Breaking Changes:**

- The `codePane` theme's properties have been changed to `wrapper` and `editor`. They could stay `pre` and `code`, but this wouldn't reflect the actual elements accurately.

**Future:**

It'd be great to add some instructions on how to add more prism components. The added text to the Readme is minimal but doesn't really say much.

It'd be great to make the playground and codepane's theme props only recognise `light` and `dark` so that it can be set to another value so the user can write/use their own themes.

**Improvements Summary:**

So while the bundle will grow, since more prism components are now included (which doesn't matter much since removing codemirror and babel-standalone for buble should've helper it a little), this gets rid of extra setup steps for the user. Previously a Prism CSS and CDN bundle would have to be added manually.

Also, this should make the syntax theme between the playground and the codepane 100% consistent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,14 +123,11 @@ We can start with this project's sample at [`one-page.html`](./one-page.html). I
     <title>Spectacle</title>
     <link href="https://fonts.googleapis.com/css?family=Lobster+Two:400,700" rel="stylesheet" type="text/css">
     <link href="https://fonts.googleapis.com/css?family=Open+Sans+Condensed:300,700" rel="stylesheet" type="text/css">
-    <link href="https://unpkg.com/prismjs@1/themes/prism-tomorrow.css" rel="stylesheet" type="text/css">
     <link href="https://unpkg.com/normalize.css@7/normalize.css" rel="stylesheet" type="text/css">
     <link href="https://unpkg.com/spectacle/lib/themes/default/index.css" rel="stylesheet" type="text/css">
 </head>
 <body>
     <div id="root"></div>
-    <script src="https://unpkg.com/prismjs@1/prism.js"></script>
-    <script src="https://unpkg.com/prismjs@1/components/prism-jsx.min.js"></script>
     <script src="https://unpkg.com/prop-types@15/prop-types.js"></script>
     <script src="https://unpkg.com/react@15/dist/react.js"></script>
     <script src="https://unpkg.com/react-dom@15/dist/react-dom.js"></script>
@@ -485,8 +482,6 @@ This tag displays a styled, highlighted code preview. I prefer putting my code s
 |---|---|---|
 |lang|PropTypes.string| Prism compatible language name. i.e: 'javascript' |
 |source| PropTypes.string| String of code to be shown |
-
-You can change your syntax highlighting theme by swapping the prism.js CSS file in `index.html`
 
 <a name="code-base"></a>
 #### Code (Base)
--- a/docs/tag-api.md
+++ b/docs/tag-api.md
@@ -112,7 +112,7 @@ This tag displays a styled, highlighted code preview. I prefer putting my code s
 |lang|PropTypes.string| Prism compatible language name. i.e: 'javascript' |
 |source| PropTypes.string| String of code to be shown |
 
-You can change your syntax highlighting theme by swapping the prism.js CSS file in `index.html`. Prism.js supports `Markup`, `CSS`, `C-like`and `Javascript` languages out of the box. To add more [supported languages](http://prismjs.com/#languages-list), you have to load the respective language's prism.js plugin in your index.html. 
+Prism.js supports `Markup`, `CSS`, `C-like`and `Javascript` languages out of the box. To add more [supported languages](http://prismjs.com/#languages-list), you have to import the respective language's prism.js component/plugin from `prismjs/components/prism-javascript`.
 
 <a name="code-base"></a>
 ### Code (Base)
--- a/docs/tag-api.md
+++ b/docs/tag-api.md
@@ -112,7 +112,7 @@ This tag displays a styled, highlighted code preview. I prefer putting my code s
 |lang|PropTypes.string| Prism compatible language name. i.e: 'javascript' |
 |source| PropTypes.string| String of code to be shown |
 
-Prism.js supports `Markup`, `CSS`, `C-like`and `Javascript` languages out of the box. To add more [supported languages](http://prismjs.com/#languages-list), you have to import the respective language's prism.js component/plugin from `prismjs/components/prism-javascript`.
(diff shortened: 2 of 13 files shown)


src/utils/prismImport.js
```diff
@@ -0,0 +1,48 @@
+/* eslint-disable filenames/match-regex */
```
ryan-roemer: Why did we disable `filenames` for this file?

src/utils/prismImport.js
```diff
@@ -0,0 +1,48 @@
...
+
+import 'prismjs/components/prism-core';
```
ryan-roemer: How much weight does this add to be part of the bundle seeing that we're including "everything"?

src/utils/prismImport.js
philpl: The rule didn’t like something about the prism core import :(

src/utils/prismImport.js
philpl: All in all if we take the other React-live PR into account it’s a net minus altogether:

> before the react-live and codepane refactors: 2.4M (min, non-gzip, without prism) and *after:* 1.6M (min, non-gzip, with prism)

src/utils/prismImport.js
ryan-roemer: It's complaining about `src/utils/prismImport.js` name of this file, not the `import`s within the code I believe.

ryan-roemer: Actually, all the _prismjs_ imported file names abide by the lint rule we have for this repo 😛

src/utils/prismImport.js
ryan-roemer: How much is just the prism part if you "external" it?

src/utils/prismImport.js
philpl: Oh, ok. Sorry, I’ll make a change

src/utils/prismImport.js
philpl: I’ll check

docs/tag-api.md
```diff
@@ -112,7 +112,7 @@ This tag displays a styled, highlighted code preview. I prefer putting my code s
 |lang|PropTypes.string| Prism compatible language name. i.e: 'javascript' |
 |source| PropTypes.string| String of code to be shown |
 
-You can change your syntax highlighting theme by swapping the prism.js CSS file in `index.html`. Prism.js supports `Markup`, `CSS`, `C-like`and `Javascript` languages out of the box. To add more [supported languages](http://prismjs.com/#languages-list), you have to load the respective language's prism.js plugin in your index.html. 
+Prism.js supports `Markup`, `CSS`, `C-like`and `Javascript` languages out of the box. To add more [supported languages](http://prismjs.com/#languages-list), you have to import the respective language's prism.js component/plugin from `prismjs/components/prism-javascript`.
```
mAAdhaTTah: It looks like you're importing a number of `prism` components, so you may want to modify this to match.

philpl: good point, I'll clarify this a little

src/utils/prismImport.js
philpl: @ryan-roemer so, it's tough to exclude prism without excluding buble as both are also imported by react-live with the respective prism subpaths as well, but if I exclude: prism, react-live, and (thus also) buble, then I we land at ~1.33MB

ryan-roemer: So that's just a `~0.3MB` min difference? If so, I'm totally cool with that...

philpl:
@carlospaelinck where did you encounter these errors? I can take a look and see whether it outputs any warnings/errors w/ this PR

carlospaelinck:
@philpl Yeah that sounds fine!

src/components/__snapshots__/code-pane.test.js.snap
```diff
@@ -13,57 +15,80 @@ exports[`<CodePane /> should render correctly. 1`] = `
...
+        undefined,
         Object {},
         undefined,
       ]
     }
   >
-    <pre
-      className="css-0"
+    <div
+      className="react-live react-live-dark  css-0"
```
ryan-roemer: NIT: Two spaces after `react-live-dark`?

README.md
```diff
@@ -123,14 +123,11 @@ We can start with this project's sample at [`one-page.html`](./one-page.html). I
     <title>Spectacle</title>
     <link href="https://fonts.googleapis.com/css?family=Lobster+Two:400,700" rel="stylesheet" type="text/css">
     <link href="https://fonts.googleapis.com/css?family=Open+Sans+Condensed:300,700" rel="stylesheet" type="text/css">
-    <link href="https://unpkg.com/prismjs@1/themes/prism-tomorrow.css" rel="stylesheet" type="text/css">
     <link href="https://unpkg.com/normalize.css@7/normalize.css" rel="stylesheet" type="text/css">
     <link href="https://unpkg.com/spectacle/lib/themes/default/index.css" rel="stylesheet" type="text/css">
```
ryan-roemer: Can we update the spectacle unpkg link to include a major semver range? Here we'll proactively get ahead of things with our anticipated major bump to `4`:

```
    <link href="https://unpkg.com/spectacle@^4/lib/themes/default/index.css" rel="stylesheet" type="text/css">
...
    <script src="https://unpkg.com/spectacle@^4/dist/spectacle.js"></script>
    <script src="https://unpkg.com/spectacle@^4/lib/one-page.js"></script>
```

philpl:
@carlospaelinck I've got a follow up PR for the props issue here: https://github.com/FormidableLabs/spectacle/pull/391 (includes a fix for the key propagation)

README.md
philpl: Done :) https://github.com/FormidableLabs/spectacle/pull/389/commits/c190c00460a247b679d3b8be2a701a8c7a509ddb

The links will obviously not work until we release v4

Issue #375, opened 2017-10-19
Add Handler to specific slide


kwelch:
This would allow for the ability to do a 'choose your own presentation' model. 

Created from conversations on [this thread](https://twitter.com/ken_wheeler/status/921038046853763077)

kwelch:
I reviewed the PR and I see how you would be able to say that Slide 2 goes to Slide 5 with the `goTo` prop.

How would one setup a button on a slide that would allow you to go to either Slide 5 or Slide 3?

I apologize because I didn't describe this functionality well. 

Here is my attempt of a graph of what I am hoping to be able to do. 

```
      1
      |
      2
     /    \
    3     4
   / |      | \
  5 6    5 7
```

carlospaelinck:
Yeah, we’re going to add support for a button that does effectively calls goTo and supports the slide number or identifier.

kwelch:
Awesome! Thanks.

carlospaelinck:
@kwelch Hopefully `<GoToAction />` gives you what you need. Two configs supported, standard button or custom rendered component.

kwelch:
That is epic! Nice works. Thank you so much for this. Cannot wait to use it.

PR #390, opened 2017-11-01
Added out-of-order slide navigation.

carlospaelinck:
- [x] Adds support for out-of-order presenting via a `goTo` prop on `<Slide />`
- [x] Added unit tests for helper functions for calculating offsets and next unviewed slide index

I used a `Set` to track viewed indexes to ensure slides are only shown once. Indexes _are_ removed when we navigate backwards so slides can be re-shown. A viewed index ordered set is necessary to prevent an infinite loop. For example, if Slide 2 directs to 4, and Slide 4 directs to 3, when we go forward from 3, we want to go to 5, otherwise we will be stuck in a loop of 4 to 2, 2 to 3 to 4.

Addresses #375

### DIFF
--- a/src/components/manager.js
+++ b/src/components/manager.js
@@ -111,7 +111,7 @@ export class Manager extends Component {
       slideReference: [],
       fullscreen: window.innerHeight === screen.height,
       mobile: window.innerWidth < props.contentWidth,
-      autoplaying: props.autoplay,
+      autoplaying: props.autoplay
     };
   }
 
@@ -148,6 +148,9 @@ export class Manager extends Component {
   componentWillUnmount() {
     this._detachEvents();
   }
+
+  viewedIndexes = new Set();
+
   _attachEvents() {
     window.addEventListener('storage', this._goToSlide);
     window.addEventListener('keydown', this._handleKeyPress);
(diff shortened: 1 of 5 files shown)


src/components/manager.js
```diff
@@ -148,6 +148,9 @@ export class Manager extends Component {
   componentWillUnmount() {
     this._detachEvents();
   }
+
+  viewedIndexes = new Set();
```
carlospaelinck: `viewedIndexes` is just a property on the class because we don’t need to re-render the Manager when it changes. Therefore having it in state isn't necessary.

src/components/manager.js
```diff
@@ -314,6 +318,27 @@ export class Manager extends Component {
...
+  _getOffset(slideIndex) {
+    const { goTo } = this.state.slideReference[slideIndex];
+    const nextUnviewedIndex = this._nextUnviewedIndex();
+    if (goTo && !isNaN(parseInt(goTo))) {
+      const goToIndex = () => {
+        if (this.viewedIndexes.has(goTo - 1)) {
+          return this._nextUnviewedIndex();
+        }
+        return goTo - 1;
+      };
```
carlospaelinck: The `goToIndex()` func prevents the infinite viewing loop. If we’ve already seen that out-of-order slide, let’s go to the next unviewed index.

PR #392, opened 2017-11-01
Add fullscreen button to ComponentPlayground

### LINKED ISSUE #316: Full Screen ComponentPlayground
It would be really nice when a slide comes up with the ComponentPlayground in it to be able to push the playground to fill the entire window. 

Depending on the resolution of the projector that I'm using the playground can be very small. If there were a button on the header that would allow it to be expanded and contracted that would make live coding in a presentation amazing.

I'm happy to take this on and submit a PR for it, just not sure if its already on the roadmap or not.

philpl:
Fix #316 
Depends on #391 

Adds a fullscreen button to the `ComponentPlayground` that uses the Fullscreen API.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "normalize.css": "^7.0.0",
     "prismjs": "^1.6.0",
     "react-emotion": "^8.0.8",
-    "react-live": "^1.8.0-1",
+    "react-live": "^1.8.0-2",
     "react-redux": "^5.0.5",
     "react-transition-group": "^1.1.3",
     "react-typography": "^0.16.5",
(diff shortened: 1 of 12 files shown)


philpl:
Quick note after testing Safari w/ @carlospaelinck: this doesn’t take into account yet that spectacle might already be in full screen mode

ebertmi:
@philpl I implemented a custom "editor" component and also integrated a fullscreen mode by putting the content into a wrapper with absolute positioning and a z-index. This works even if spectacle is already in fullscreen mode.

philpl:
@evertmi the way to go if we need more fullscreen-able components is definitely an abstraction, maybe even a portal component, to support Safari.

For now this should definitely be sufficient for the current use case

PR #394, opened 2017-11-02
Add GoToSlide button

carlospaelinck:
- [x] Add GoToSlide button that lets you jump to a slide via number or string id
- [x] Lets you do a simple button that extends `Base` styles
- [x] Also has a render prop for custom rendering that passes the function
- [x] Renders just a div if neither a slide id or render prop is provided
- [x] Added docs with 2 examples

Finishes off #375 

screen shot 2017-11-02 at 4 33 08 pm
screen shot 2017-11-02 at 4 33 14 pm


kenwheeler:
👍

PR #405, opened 2017-11-08
Magic move

kenwheeler:
Initial run at magic move, will have separate PRs for docs and keying features.


src/components/magic-wrapper.js
```diff
@@ -0,0 +1,223 @@
...
+        <Deck>{this.props.children}</Deck>
+      </Context>,
+      this.portal,
+      () => {
+        this.timeout = setTimeout(() => {
+          updateChildren(this.container.childNodes[0]);
+          updateChildren(this.portal.childNodes[0].childNodes[0]);
+          buildStyleMap(
+            this.portalMap,
+            this.portal.childNodes[0].childNodes[0]
```
carlospaelinck: Will the first element of these arrays always exist?

kenwheeler: Yes, but i'll code defensively against it anyway

PR #407, opened 2017-11-09
Sync the component-playground across tabs via local storage

philpl:
Because live coding... duh 😛

### DIFF
--- a/src/components/component-playground.test.js
+++ b/src/components/component-playground.test.js
@@ -3,6 +3,14 @@ import { render, shallow } from 'enzyme';
 import ComponentPlayground, { PlaygroundProvider } from './component-playground';
 
 describe('<ComponentPlayground />', () => {
+  beforeEach(() => {
+    window.localStorage = { setItem: () => {} };
+  });
+
+  afterEach(() => {
+    window.localStorage = undefined;
+  });
+
   const context = { styles: {
     components: { syntax: {} },
     prism: { light: 'light;', dark: 'dark;' }
--- a/src/components/component-playground.test.js
+++ b/src/components/component-playground.test.js
@@ -2,13 +2,15 @@ import React from 'react';
 import { render, shallow } from 'enzyme';
 import ComponentPlayground, { PlaygroundProvider } from './component-playground';
 
+const origLocalStorage = window.localStorage;
+
 describe('<ComponentPlayground />', () => {
-  beforeEach(() => {
+  beforeAll(() => {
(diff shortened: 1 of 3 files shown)


src/components/component-playground.test.js
```diff
@@ -3,6 +3,14 @@ import { render, shallow } from 'enzyme';
 import ComponentPlayground, { PlaygroundProvider } from './component-playground';
 
 describe('<ComponentPlayground />', () => {
+  beforeEach(() => {
+    window.localStorage = { setItem: () => {} };
```
ryan-roemer: Let's not permanently destroy `localstorage` for the other tests.

suggest at the top:

```
const realStorage = localStorage;

// ...

afterEach((() => {
  window.localStorage = realStorage;
});
```

ryan-roemer: Actually, if `setItem()` is all you need to override, just `sandbox.stub()` it.

philpl: @ryan-roemer I just copied this from the manager's tests :wink: I think we're testing in an environment where the local storage is not mocked out-of-the-box anyway

ryan-roemer: I'm not totally following. The `window.localStorage = undefined;` is a potential footgun down the road for unrelated tests that may use `localStorage` for something innocuous...

ryan-roemer: If we're super sure of that, then as a protective measure, I suggest a global `beforeEach` that sets `window.localStorage = undefined;` so we discover any kinks immediately in the future...

philpl: Depends; we'd have to update it here as well: https://github.com/FormidableLabs/spectacle/blob/master/src/components/manager.test.js#L65

ryan-roemer: Let's do that. It's a footgun here and in the existing code, so we'll just take care of both.

philpl: @ryan-roemer Is this what you had in mind? https://github.com/FormidableLabs/spectacle/pull/407/commits/af1c727b0af1b4a0055f2b5d5407bdad82f8214c

ryan-roemer: Yep -- or maybe move `const origLocalStorage = window.localStorage;` above the first `describe`, so that you're sure no tests have mucked with things. And you only have to set it once. But your call there -- as you have it already is fine too...

philpl: Good point!

PR #410, opened 2017-11-10
Ignore non-DOM Editor props

philpl:
- Filter style props that are only used for Emotion in the Editor
  components
- Edit ESLint no-unused-vars rule to ignore /^_+$/ variables

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -47,6 +47,9 @@
     "react/jsx-handler-names": "off",
     "react/no-find-dom-node": "off",
     "no-invalid-this": "off",
-    "complexity": "off"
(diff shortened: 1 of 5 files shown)


.eslintrc
```diff
@@ -47,6 +47,9 @@
     "react/jsx-handler-names": "off",
     "react/no-find-dom-node": "off",
     "no-invalid-this": "off",
-    "complexity": "off"
+    "complexity": "off",
+    "no-unused-vars": ["error", {
+      "argsIgnorePattern": "^_+$"
```
carlospaelinck: 👍

Issue #329, opened 2017-06-09
Unable to Use Markdown containing code.


jtama:
Given the following markdown file : 
````javascript
console.log('' == '0');
console.log(0 == '');
console.log(0 == '0');
console.log(false == 'false');
console.log(false == '0');
console.log(false == undefined);
console.log(false == null);
console.log(null == undefined);
console.log(' \t\r\n' == 0);
````
And the following slide : 
`
import equalsSlide from "../assets/equals.md";
...

                <Slide>
                    <Markdown source={equalsSlide}/>
                </Slide>
`

The rendered presentation triggers the following error.

`
marked.js:138 Uncaught TypeError: Cannot read property 'replace' of undefined
Please report this to https://github.com/chjj/marked.
    at Lexer.lex (marked.js:138)
    at Function.Lexer.lex (marked.js:129)
    at marked (marked.js:1219)
    at exec (index.js:184)
    at Markdown.render (markdown.js:143)
    at ReactCompositeComponent.js:796
    at measureLifeCyclePerf (ReactCompositeComponent.js:75)
    at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (ReactCompositeComponent.js:795)
    at ReactCompositeComponentWrapper._renderValidatedComponent (ReactCompositeComponent.js:822)
    at ReactCompositeComponentWrapper.performInitialMount (ReactCompositeComponent.js:362)
`

**Comment by mAAdhaTTah** [comment]
I just ran into this with an inline code bit:

```markdown
Here is a piece of `code`
```

Slightly different stack, but same error:

```
    at format (webpack-internal:///./node_modules/spectacle/lib/components/code-pane.js:47)
    at RadiumEnhancer.createMarkup (webpack-internal:///./node_modules/spectacle/lib/components/code-pane.js:69)
    at RadiumEnhancer.render (webpack-internal:///./node_modules/spectacle/lib/components/code-pane.js:80)
    at RadiumEnhancer.render (webpack-internal:///./node_modules/radium/lib/enhancer.js:132)
    at eval (webpack-internal:///./node_modules/react-dom/lib/ReactCompositeComponent.js:795)
    at measureLifeCyclePerf (webpack-internal:///./node_modules/react-dom/lib/ReactCompositeComponent.js:75)
[...]

PR #427, opened 2017-11-19
Inline code w/ markdown produces a `CodePane`, not `Code`

mAAdhaTTah:
Created a test case for it. You can see in the snapshot that a `CodePane` is produce, which should be a `Code` element.

Both the `Code` and `CodePane` elements work strangely when produced by the Markdown output. Besides this, correctly formatted `CodePane` elements are only highlighted if the slide is focused on page load. Switching into a Slide with a `CodePane` generated by Markdown leaves the `CodePane` unhighlighted. Pretty strange.

This was also an issue in Spectacle 2.* (which I was using previously). I'm in the process of upgrading to 4.0, which also means converting the Markdown slides into React ones, because I keep running into this issue and need more control over the styles.

I'm opening this PR to get some thoughts as to why these things may be happening and where I should look to fix it.

### DIFF
--- a/example/src/index.js
+++ b/example/src/index.js
@@ -203,6 +203,7 @@ export default class Presentation extends React.Component {
   * Lists too!
   * With ~~strikethrough~~ and _italic_
   * And let's not forget **bold**
+  * Add some \`inline code\` to your sldes!
             `}
           </Markdown>
         </Slide>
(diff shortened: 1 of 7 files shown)


mAAdhaTTah:
Digging into this, I'm beginning to think this is a bug in marksy itself. It doesn't appear `codespan` is even making an appearance the way we expect it to.

mAAdhaTTah:
Yeah, error is in marksy, where both inline codespans & code blocks are route through the same element. PR w/ marksy test case is here: https://github.com/FormidableLabs/spectacle/pull/427 probs not going to get a chance to investigate much deeper than that today, but maybe the marksy guys can figure it out.

mAAdhaTTah:
The latest version of marksy resolves the issue noted here. I also borrowed the code I wrote for #337, which resolves the `replace` error reported #329 and either reintroduced or never resolved in Spectacle 4.*. The "CodePane not highlighting" issue mentioned above is also not an issue, so that may have been a 2.* only issue (which I was using previously). This PR resolves all the markdown+code-related issues in spectacle currently.

I generally prototype my slides in Markdown before moving them over to React components as needed, so getting this merged and pubbed would be extremely beneficial to my workflow.

Let me know if you have any suggestions. Specifically, the `this.props.children` appeared as an array when used with marksy, so you may have an alternate approach preferred for dealing with that.

Thanks for the great project!

PR #432, opened 2017-11-21
Fixes #422 (Magic Move)

### LINKED ISSUE #422: Magic Move doesn't align text properly
In the announcement tweet, https://twitter.com/FormidableLabs/status/930457800832094209, it's not quite a smooth transition. You can tell that the text doesn't match up at the beginning of the transition. I did a screen recording and stepped through the frames so you can really see it.

Unfortunately github doesn't support .mov files so you'll have to download it: http://jlongster.com/s/magic-move.mov

I haven't actually used it yet, just noticed it in that video, so I'm not sure if it happens in all cases or not.

(btw this is a very cool feature)

kenwheeler:
This fixes 2 things:

- Takes into account the positioning of the parent elements
- Repurposes already prism-wrapped copy inside of spans

### DIFF
--- a/src/utils/magic.js
+++ b/src/utils/magic.js
@@ -1,5 +1,5 @@
 /* eslint-disable max-statements */
-function wrapTextChildren(child, valMap) {
+function wrapTextChildren(child, valMap, classes) {
   const frag = document.createDocumentFragment();
   child.textContent.split('').map(s => {
     let char = s;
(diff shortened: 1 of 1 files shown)


src/utils/magic.js
```diff
@@ -59,6 +62,10 @@ export function updateChildren(root, valMap = {}) {
     }
 
     if (child.nodeType === 1) {
+      if (child.parentNode.classList.contains('prism-code') && child.tagName === 'SPAN') {
```
philpl: Nice :heart:

PR #446, opened 2017-12-29
Add bgLighten to Base

### LINKED ISSUE #445: [FR] Add bgLighten config/prop
Add a prop to enable `bgLighten` like `bgDarken` for `bgImage`.

aravindballa:
Fixes #445 

Added a new prop `bgLighten` to Base so that it supports lightening images.

Screenshot:

Screenshot_5

### DIFF
--- a/src/utils/base.js
+++ b/src/utils/base.js
@@ -51,6 +51,15 @@ const convertFontSizeToPx = function (fontSize) {
   return convertedFontSize;
 };
 
+const hexToRGB = function (hex) {
+  hex = (hex.charAt(0) === '#') ? hex.substring(1, 7) : hex;
+  const RGB = [];
+  RGB.push(parseInt(hex.substring(0, 2), 16)); // R
+  RGB.push(parseInt(hex.substring(2, 4), 16)); // G
+  RGB.push(parseInt(hex.substring(4, 6), 16)); // B
+  return RGB;
+};
+
 export const getStyles = function getStyles() {
   if (process.env.NODE_ENV !== 'production' && typeof this.warnedAboutFontSize === 'undefined') {
     this.warnedAboutFontSize = false;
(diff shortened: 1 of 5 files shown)


aravindballa:
Custom overlay colours would be a great one. Something like `bgOverlayColour` and `bgOverlayIntensity`.

kenwheeler:
Lets do that. That will be tight

aravindballa:
Please review and check whether we can continue like this or not.
@kenwheeler @philpl

src/utils/base.js
```diff
@@ -51,6 +51,15 @@ const convertFontSizeToPx = function (fontSize) {
   return convertedFontSize;
 };
 
+const hexToRGB = function (hex) {
```
philpl: Instead of maintaining this piece of code ourselves it should probably use a helper. Polished has some excellent ones and we won't need to pull in the entire library ✨ 

https://polished.js.org/docs/#transparentize
  
or

https://polished.js.org/docs/#parsetorgb

src/utils/base.js
```diff
@@ -68,7 +77,8 @@ export const getStyles = function getStyles() {
     textAlign,
     bgColor,
     bgImage,
-    bgDarken,
+    bgOverlayColour,
+    bgOverlayIntensity,
```
philpl: I wouldn't call this prop "intensity" as that would refer to the saturation of the colour: https://www.uwgb.edu/heuerc/2D/ColorTerms.html

Instead I would just call this "opacity" or "opacify" or "transparentize" depending on what it'll do.

aravindballa: Its bgTransparentize now ✨

aravindballa:
Can you verify if I did it right? @philpl 
cc @kenwheeler

aravindballa:
@kitten @kenwheeler Can you check?

ebrillhart:
Hey @aravindballa thank you so much for your work on this! I am closing this since the functionality was added in #531 (along with some other changes we needed), which will be merged soon.

PR #459, opened 2018-01-22
[docs] Warns users not to use a transition prop w/ Magic Move

peggyrayzis:
- Added sentence warning users not to use the `transition` prop w/ Magic Move (I learned this the hard way 😅 )


kenwheeler:
Nice! and sorry 🙃

PR #460, opened 2018-01-23
Add `theme="external"` to disable the built-in prism themes

### LINKED ISSUE #458: CodePane doesn't respect custom Prism theme
Hi! I'm using the most recent version of Spectacle via CRA. Unfortunately, CodePane isn't respecting my custom Prism theme.

I've tried:
1. Adding `<link rel="stylesheet" type="text/css" href="https://cdn.jsdelivr.net/npm/prismjs@1.10.0/themes/prism-okaidia.css">` to the <head> of my `index.html` file
2. Importing the theme in my `presentation.js` file: `import 'prismjs/themes/prism-okaidia.css';`

In Chrome dev tools, you can see the css for the theme has loaded, but it's being overridden by whatever theme Spectacle includes by default.

Thanks for your help!

kitten:
Fix #458

When applied this PR 'hides' our built-in Prism themes behind the `.builtin-prism-theme` class, which is applied by default. It can be disabled by setting the `theme` prop to `"external"` which allows external themes to style Prism in whatever way they'd like.

This also adds a stricter prop-types check for the theme prop, and passes on the correct `language-X` class for Prism. (Quick on-the-side-fix)

A user might still run into specificity problems where themes can mix when their own Prism theme contains some really generic selectors with a higher specificity than ours. I don't think there's any value in preventing this with even more specific selectors on our end, instead the README now just mentions this.

~Edit: not quite sure why, but an unrelated snapshot changed as well. Might be worth to take a look at that as well :/~
Edit 2: nevermind, I forgot to run yarn 🤦‍♂️

### DIFF
--- a/README.md
+++ b/README.md
@@ -573,8 +573,15 @@ This tag displays a styled, highlighted code preview. I prefer putting my code s
 |---|---|---|
 |lang|PropTypes.string| Prism compatible language name. i.e: 'javascript' |
 |source| PropTypes.string| String of code to be shown |
+|className| PropTypes.string| String of a className to be appended to the CodePane |
+|theme| PropTypes.string| Accepts `light`, `dark`, or `external` for the source editor's syntax highlighting. Defaults to `dark`. |
 
-If you want to change the theme used here, you can include a prism theme in index.html via a script tag. CodePane and Playground both use the prism library under the hood, which has several themes that are available to include.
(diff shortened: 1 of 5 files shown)


kitten:
@kenwheeler I just went ahead and merged this since it was pending for a while. If anyone else finds sth wrong with th it we can always path or revert :)

PR #468, opened 2018-02-10
Update ComponentPlayground docs (default theme)

jonasws:
The README initially gave `light` as the default theme for a `CompoonentPlayground`, while according to the source for said component, the default theme is indeed `dark`.


jonasws:
Looks to have been fixed. Closing.

PR #471, opened 2018-02-12
Add aria-label to fullscreen button

alexlande:
Adds an `aria-label` to the fullscreen button so that screen reader users can tell what it is.


kenwheeler:
Thanks dude!

PR #477, opened 2018-02-22
Add protocol to emotion link

btmills:
Previously, the browser treated `github.com/emotion-js/emotion` as a relative link. Specifying `https://` gives the browser a full URL so it correctly navigates to the Emotion repository.


ryan-roemer:
Thanks!

PR #479, opened 2018-02-24
babel-preset-es2015 -> babel-preset-env

alirezavalizade:


ryan-roemer:
Thanks for the PR -- we've switched to env w/ babel7 in #493

PR #483, opened 2018-03-06
Fancier <Appear> animations

elifitch:
spectacle-anim-gif

First, thanks so much for your hard work building Spectacle. I'm *really* enjoying it so far.

Anyhoo I'm using spectacle to build a conference talk on animation, and I needed some more robust animation options for the Appear component. They were really quick to implement and don't change the default behavior of Appears so I figured I'd PR em back in to the project. It's a tiny change but you can use it to make some really neat effects 😊

### DIFF
--- a/src/components/appear.js
+++ b/src/components/appear.js
@@ -59,19 +59,26 @@ class Appear extends Component {
   }
 
   render() {
-    const child = React.Children.only(this.props.children);
-    const endValue = this.state.active ? 1 : 0;
-    const transitionDuration = this.props.transitionDuration;
+    const {
+      children,
+      startValue,
+      endValue,
+      transitionDuration,
+      easing,
+      style
+    } = this.props;
+    const child = React.Children.only(children);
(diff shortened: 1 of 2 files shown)


src/components/appear.js
```diff
@@ -60,18 +60,18 @@ class Appear extends Component {
 
   render() {
     const child = React.Children.only(this.props.children);
-    const endValue = this.state.active ? 1 : 0;
+    const tweenData = this.state.active ? this.props.endValue : this.props.startValue;
```
kyleshevlin: I would be a fan of destructuring `props` to clean this up more. Above line 62, you could destructure `children`, `easing`, `endValue`, and `startValue`. Otherwise, I think this is solid addition, but I'll leave that to the requested reviewers to decide.

elifitch:
@kyleshevlin destructured those props 😄

kenwheeler:
Whoah this is sick

Issue #488, opened 2018-03-17
<Appear /> + Emotion styled component = many style tags


elifitch:
Every time an `<Appear />` tag executes a tween, it looks like emotion is adding a bunch of style tags to the head.

![mar-17-2018 17-38-30](https://user-images.githubusercontent.com/2495933/37560172-16d3a20a-2a0a-11e8-8c08-2cc71e19c553.gif)

Wrapping the children of an `<Appear />` in a native dom element seems to be a pretty solid workaround, but should probably be noted somewhere in the documentation until this issue is resolved.

I'm working on a PR to Spectacle which would allow more flexible animations, and I'd be happy to add a note to the docs along with my PR if you'd like.

kitten:
I think the problem here is quite simple. Spectacle's Appear component is right now just passing on a `style` prop;

https://github.com/FormidableLabs/spectacle/blob/master/src/components/appear.js#L81

Now unfortunately some components take that `style` prop and are interpolating it into emotion, which then considers it part of its styles, which means that it'll be mounted into the style tag. 

https://github.com/FormidableLabs/spectacle/blob/master/src/components/text.js#L99-L103

Now, this is not a horrible thing to do, but it'll definitely constrict the performance, as it'd be much faster to in this case just use the normal React `style` prop, which I'd consider a best practice for both emotion and styled-components.

cc @kenwheeler: Any thoughts? Would this be a pain to change?

kenwheeler:
Does this happen in prod as well as dev?

kale-stew:
Closing for lack of response from the OP

PR #492, opened 2018-03-18
Flexible <Anim> component

elifitch:
mar-17-2018 18-14-29

**Hi again Formidable folks 👋!** Really loving my time with spectacle, it's super fun to work with and a nice change of pace from reveal.js. Super appreciate all the work y'all've done on it and the opportunity to contribute to the project.

I'm using Spectacle to make a deck for a talk on web animation, so I wanted some more flexibility animating fragments. Specifically I wanted to be able to animate a fragment multiple times before moving on to the next fragment or slide.  I ended up more-or-less replacing the `<Appear>` component with a more generalized `<Anim>` component, creating a simpler `<Appear>` component which is just a wrapper for `<Anim>` to avoid shipping any breaking changes, and made a few small alterations in `<Slide>` and `<Manager>` to accommodate the ability to step through the same fragment several times.

You could use this to make a little carousel, step through an animated infographic, and so many other things my small brain can't conceive of. These changes also make animation in Spectacle much more extensible. They've enabled me to make a really robust GSAP powered animation component for my talk that wouldn't have been possible otherwise.

I concede that I would have liked to add some tests to lock down the new fragment logic, but I had some trouble grokking the tests for `<Manager>` and figuring out how to get the fragment functionality tested there.

I left my commits raw just in case you prefer looking at a PR that way, but I'd be totally happy to squash if that would help.

Oh, also went ahead and included documentation on the workaround discussed in #488.

### DIFF
--- a/README.md
+++ b/README.md
@@ -571,7 +571,7 @@ For best performance, wrap the contents of this tag in a native DOM element like
 |fromStyle|Proptypes.object|A style object that defines the starting, inactive state of the Anim tag.
 |toStyle|Proptypes.array|An array of style objects that define each step in the animation. They will step from one toStyle object to another, until that fragment is finished with its animations.
 |easing|PropTypes.string|A victory easing curve for the Appear animation. The various options are documented in the [Victory Animation easing docs](https://formidable.com/open-source/victory/docs/victory-animation/#easing).
-
+|onAnim|PropTypes.fun|This function is called every time the Anim component plays an animation. It'll be called with two arguments, forwards, a boolean indicating if it was stepped forwards or backwards, and the index of the animation that was just played.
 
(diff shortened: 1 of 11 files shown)


kenwheeler:
is this good to go?

elifitch:
@kenwheeler as far as I'm concerned yeah I think so, as long as you're okay with the testing situation. I would have preferred to add some tests for the new fragment behavior, but I wasn't 100% sure whether the manager tests was the place to add them, and I had some trouble wrapping my head around the tests there.

elifitch:
@kenwheeler anything I can do to carry this across the finish line?

kenwheeler:
Just had to find some time to manually test it. 💯

PR #498, opened 2018-03-20
Infra: Upgrade to webpack@4

### LINKED ISSUE #494: Infra: Update webpack build.
- [x] Add `webpack@4`.
- [x] Refactor all configs, check loaders still work.
- [x] Check status of `hot`, `hmr` things that we want.
- [x] Try to use simple `mode`s
- [x] Check on `node server.js` / `yarn start`

ryan-roemer:
- Upgrade to `webpack@4` and related packages. Fixes #494
- Minor eslint harmonization and package updates.
- Replace `webpack-dev-middleware` + `webpack-hot-middleware` + `express` with `webpack-dev-server`.
    - _Note_: We now get page refreshes to a single slide that changes, but it does rekey on the right slide, so I think that's good enough, especially considering the reduction in complexity from just using wds...
- Hone down and harmonize webpack configs (for future prettier work).
- Hone down `.npmignore` more. We were erroneously publishing `dist/bundle.js`, the **example** slide deck and other stuff.

/cc @kenwheeler @carlospaelinck


.npmignore
```diff
@@ -1,7 +1,7 @@
 /*
+/dist/*
 !/dist/spectacle*
-dist/*.map
-!/docs
+/dist/*.map
```
ryan-roemer: We were publishing `dist/bundle.js` and other example deck stuff. this hones it down more, and removes `docs` because they're not needed for downloads to save space. Current `ls`:

```
LICENSE
README.md
dist
es
lib
package.json
src
```

ryan-roemer:
Released in `spectacle@4.2.1`

Issue #496, opened 2018-03-20
Infra: Add prettier, tighten lint.


ryan-roemer:
- [ ] Add prettier
- [ ] Tighten formatting (via lint or prettier) for better consistency.
- [ ] Maybe add precommit hooks.

joshuacrass:
@ryan-roemer I'm looking to get involved in some open source projects. Do you mind if I work on this?

PR #511, opened 2018-04-11
Infra/496/add prettier

joshuacrass:
Added prettier and pre-commit hooks for linting and prettier (#496) 

The checks will only run on staged files. This will only "check" the files and not fix them. The commit will fail if any staged files do not pass.  

If you need/want to fix them, you will need to run `yarn run prettier --write <filename>`.

There are a few options to prettier your files. I would recommend, if possible, to add the prettier plugin to your editor.

Note that as a result of editing eslintrc and package.json, there are several whitespace changes. We must also keep in mind that editing any file that does not currently conform to prettier standards will also be affected by several formatting changes.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+package-lock.json
+package.json
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,11 @@
 {
   "root": true,
   "parser": "babel-eslint",
-  "extends": "formidable/configurations/es6-react",
+  "extends": [
+    "formidable/configurations/es6-react",
+    "prettier",
+    "prettier/react"
+  ],
   "env": {
     "browser": true,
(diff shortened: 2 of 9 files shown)


.prettierignore
```diff
@@ -0,0 +1 @@
+package-lock.json
```
ryan-roemer: Let's also remove `package.json`. `yarn` and `npm` write these in incompatible ways IIRC.

joshuacrass: ✅

sxzhao:
@joshuacrass this looks good! Please fix the merge conflicts. If all looks good I’ll go ahead and merge.

zhaostan:
@joshuacrass thanks for fixing the merge conflicts! Looks like now CI is failing, if you sync with master it should fix it. Thanks!

PR #517, opened 2018-04-25
Allow slide margin to be 0

JakeGinnivan:
0 is falsy so it defaults to 40 unless you use the string 0 or `0px`.


kenwheeler:
Can you update the snapshots?

kenwheeler:
Whoops, looks like lint too

JakeGinnivan:
@ebrillhart sorry about that, must have missed the notifications from Ken at the time

PR #528, opened 2018-05-17
Fix code pane languages

### LINKED ISSUE #526: `CodePane` throws when using `lang="java"`
Repository to reproduce: https://github.com/rainboxx/spectacle-issue

It seems that loading all language files, as of now, doesn't have any impact because `react-live` [loads Prism itself and only 4 language](https://github.com/FormidableLabs/react-live/blob/d71d13d079fc99cafcb65ec19ec057fc54a3492a/src/utils/prism.js.) files. The language files ([see Java for example](https://github.com/PrismJS/prism/blob/master/components/prism-java.js)) itself expects a global called `Prism`, so I interpret that reloading Prism removes all previously loaded language files.

rainboxx:
This PR changes the import order and overwriting of Prism, fixes #526.

Prettier changes are committed separately since they're not part of the initial change.

### DIFF
--- a/src/utils/prism-import.js
+++ b/src/utils/prism-import.js
@@ -1,5 +1,3 @@
-import 'prismjs/components/prism-core';
-
 // NOTE: This includes a "slim" list of some components of prism, but not all
 // For more, see: https://github.com/PrismJS/prism/blob/1fd690dd8a652dee375acb78b6034c3fc2db687c/components.js
 // and: https://github.com/PrismJS/prism/tree/gh-pages/components
--- a/src/components/__snapshots__/code-pane.test.js.snap
+++ b/src/components/__snapshots__/code-pane.test.js.snap
@@ -51,7 +51,7 @@ exports[`<CodePane /> should render correctly. 1`] = `
         syntaxStyles={Object {}}
       >
         <Component
-          className="language-jsx builtin-prism-theme  css-3w3gcf"
+          className="language-jsx builtin-prism-theme  css-1rm1mht"
           code="
       const myButton = (
         <CustomButton
@@ -71,7 +71,7 @@ exports[`<CodePane /> should render correctly. 1`] = `
           syntaxStyles={Object {}}
         >
           <Editor
-            className="language-jsx builtin-prism-theme  css-3w3gcf"
+            className="language-jsx builtin-prism-theme  css-1rm1mht"
             code="
       const myButton = (
         <CustomButton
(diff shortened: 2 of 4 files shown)


rainboxx:
This PR will pass CI checks once #527 is merged and this branch is rebase.

rainboxx:
#527 is merged, rebased.

rainboxx:
Missed updating snapshots, should be fixed now.

src/utils/prism-import.js
```diff
@@ -1,5 +1,3 @@
-import 'prismjs/components/prism-core';
```
samwhale: Why are we removing prism-core? This is used for the playground via react-live

ryan-roemer:
@samwhale -- We have a broader discussion of Prism imports (and conflicts with `react-live`) going on at: https://github.com/FormidableLabs/spectacle/issues/546

I suspect that ticket subsumes this specific work. Perhaps we can add a tag to collect all the prism import related issues + PRs and tackle them all at once?

/cc @kenwheeler @kitten

src/utils/prism-import.js
rainboxx: AFAIRemember it's removed since importing this overwrites the instance of `Prism` in `window`, removing all configurations made by `react-live`. It's imported by `react-live` as well, which is why I came to the conclusion that the import is no longer needed. (And it worked, AFAIRemember.)

ryan-roemer: I'm guessing piggybacking off a separate library's dependency is not a good long term approach and we really need to solve the underlying prismjs version conflict problem discussed in detail in #546 first...

/cc @kenwheeler @kitten

rainboxx: I agree but this was the solution I could find for #526. Maybe it would still be a valid approach to merge this and refactor this through #546?

ryan-roemer: Are we completely confident relying on the load order of `react-live` and then piggy-backing here won't break? Including if `react-live` changes semver under the hood what _it_ does with `prism`? I'm guessing we can't make that guarantee...

ryan-roemer: Basically, I don't want to open up _new_ bugs for a hack here when we have a larger overall issue that needs solving 😄

rainboxx: Load order: I'm confident, there's no asynchronously happening.

Semver: we can never guarantee that changes in `react-live` don't break anything here. Currently it already *is* broken since no languages other than the default of `react-live` can be used. The issue here is with how `prism` works: it creates a global object and syntax settings are added to this. Re-importing `prism` will overwrite the global object, thus removing all prior settings. This also means: importing `prism-import.js` has no effect _at all_ the way it is now (without the PR).

rainboxx:
FYI: I'm no longer using this library for myself/my project. I wanted to raise the question whether this PR is still needed/wanted given that #546 is merged.

ebrillhart:
hey @rainboxx I'm going to close this PR for the reason you mentioned - #546 should cover the major issue.

PR #535, opened 2018-06-02
[WIP]: Add bgImageStyle prop, refactor src/util/base.js 

### LINKED ISSUE #428: bgImage prop should also support gradients
I was working on a slide deck when I realized that the `bgImage` prop assumes that the value will be a path to be used with the `url()` function. In many cases, this is a nice assumption, but `background-image` also happens to be the CSS property that gradients are added with.

I would like to add support for being able to add gradients to slides. I think this can be done one of two ways.

1) We modify `bgImage` prop to expect a string without automatically wrapping the string in a `url()` call. The challenge with this approach is coming up with a way to handle the use case where a user may purposefully, though more likely inadvertently, use the `bgDarken` prop as well. We could ignore `bgDarken` if the string supplied to `bgImage` has a gradient in it.

2) Another option is to add a new prop, `bgGradient` that handles this one use case. Challenges with this will be deciding how to handle situations where all of the props (`bgImage`, `bgDarken`, and `bgGradient`) are simultaneously present, but we could come to a decision on that. This might be the cleaner of the two implementations.

I'm open to other ideas as well and happy to do the work of the PR, but wanted to get some input before going down one path or another.

DDunc:
Expected functionality when complete:
*  bgImageStyle allows for setting the backgroundImage property directly (resolve #428)
* Separate out the one-time dev-only warning flags/console.warns, make sure they provide useful information about what sugar props will be ignored when conflicts are present (also resolving #428).
* Tests for util/base.js (the refactor into smaller pure functions has made this easier)
* Eslint won't hate me

### DIFF
--- a/src/utils/base.js
+++ b/src/utils/base.js
@@ -1,23 +1,23 @@
 /*eslint max-statements:0,complexity:0,no-invalid-this:0*/
 
-const parseFontSize = function (fontSize) {
+const parseFontSize = function(fontSize) {
   const sizeComponents = fontSize.match(/\d*\.*\d+|\D+/g);
   const size = parseFloat(sizeComponents[0]);
   const unit = sizeComponents[1];
   return { size, unit };
 };
 
-const getFontSizeFromElement = function (element) {
-  const fontSize = window.getComputedStyle ?
-    window.getComputedStyle(element).getPropertyValue('font-size') :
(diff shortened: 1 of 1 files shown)


DDunc:
closing in favor of https://github.com/FormidableLabs/spectacle/pull/542

PR #531, opened 2018-06-01
Feature/add bg lighten

samwhale:
Fixes #445 
* Add bgLighten prop to base.js. Add bgLighten description to README.md
* Prettified README.md and src/utils/base.js
* Add console warning for bgLighten and bgDarken

PR #542, opened 2018-06-05
Move font-size and warn utils out of base, refactor conditionals, add bgImageStyle

### LINKED ISSUE #428: bgImage prop should also support gradients

DDunc:
*bgImageStyle prop allows for setting the backgroundImage property directly (resolve #428)
*bgLighten prop allows for setting opacity with white instead black  (based on #531, happy to rebase/refactor for the bgLighten warning)
*Refactor for greater purity/granularity/testability
*Disabled consistent-return for base.js. This seems like an Airbnb rule that's part of a larger discussion about expectations of implicit returns, but would add a lot of boilerplate here.
*Validated that old and new getStyles functions return the same objects/trigger the same one time warnings but some tests would still be a good idea.


src/utils/base.js
```diff
@@ -1,169 +1,155 @@
...
-    ? window.getComputedStyle(element).getPropertyValue('font-size')
-    : element.currentStyle.fontSize;
-  return fontSize ? parseFontSize(fontSize) : null;
+/*eslint max-statements:0,complexity:0,no-invalid-this:0,consistent-return:0*/
+import checkWarnings from './warn';
+// eslint-disable-next-line max-params
+export const buildStyles = (transforms, props, context, styles = {}) => {
+  return transforms.reduce((av, cv) => {
+    return { ...av, ...cv(props, context, av) };
+  }, styles);
```
ryan-roemer: Did we want to mutate the `styles` object passed into us?

src/utils/base.js
```diff
@@ -1,169 +1,155 @@
...
-  if (textAlign) {
-    styles.textAlign = textAlign;
+}
+export const transformItalic = ({ italic }) => {
+  if (typeof italic === 'boolean') {
+    return { fontStyle: italic ? 'italic' : 'normal' };
+  }
+};
+export const transformBold = ({ bold }) => {
+  if (typeof italic === 'boolean') {
```
ryan-roemer: `typeof italic`? Looks like the wrong variable referenced.

Shouldn't lint catch this?

src/utils/base.js
DDunc: Since we're using styles as an initial value and returning a new value, I don't think it would be expected behavior to have side effects.

src/utils/base.js
DDunc: I would certainly prefer it to, but eslint no-undef [skips typeof by default](https://eslint.org/docs/3.0.0/rules/no-undef#options). After declaring`"no-undef": ["error", { "typeof": true }]` in the eslintrc and running it against all the project files it doesn't look like we're using that typeof 'feature' anywhere else.

PR #495, opened 2018-03-19
changed slide sync to websocket (from localStorage), exposed to local network

WhoMeNope:
it is now possible to use a phone (connected to the same network) to switch slides

PR #543, opened 2018-06-05
Fix broken WS syncing.

ryan-roemer:
- Unwinds and removes websocket syncing from #495, going back to `localStorage`.
- Causes travis to fail on outdated / missing deps in `yarn.lock`.

/cc @kenwheeler @coopy @samwhale @zhaostan


kitten:
Maybe a reimplementation of websocket-syncing would make more sense to be similar to: https://github.com/kitten/spectacle-sync, as in being a separate package?

ryan-roemer:
@kitten -- Do we even need an implementation _in_ Spectacle or just some verbiage + a link to `spectacle-sync` in the README here?

samwhale:
I'm not seeing 502 errors anymore. Just this: `Error during service worker registration: DOMException: Only secure origins are allowed (see: https://goo.gl/Y0ZkNV)`. I think it's good to go.

ryan-roemer:
Released in `spectacle@4.3.1`

PR #545, opened 2018-06-05
Appear fixed when slide has id

### LINKED ISSUE #430: Appear breaks when slide has an id
For example, the following works:

```
        <Slide>
          <List>
            <ListItem>Item 1</ListItem>
            <Appear>
              <ListItem>Item 2</ListItem>
            </Appear>        
          </List>
        </Slide>
```

but the following doesn't:

```
        <Slide id="contents">
          <List>
            <ListItem>Item 1</ListItem>
            <Appear>
              <ListItem>Item 2</ListItem>
            </Appear>        
          </List>
        </Slide>
```

samwhale:
Fixes #430.

### DIFF
--- a/src/components/slide.js
+++ b/src/components/slide.js
@@ -64,7 +64,7 @@ class Slide extends React.PureComponent {
               addFragment({
                 className: frag.className || '',
                 slide: this.props.hash,
-                id: `${this.props.slideIndex}-${currentOrder}`,
+                id: `${this.props.hash}-${currentOrder}`,
                 animations: Array.from({ length: frag.dataset.animCount }).fill(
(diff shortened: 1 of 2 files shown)


src/components/slide.js
```diff
@@ -64,7 +64,7 @@ class Slide extends React.PureComponent {
               addFragment({
                 className: frag.className || '',
                 slide: this.props.hash,
-                id: `${this.props.slideIndex}-${currentOrder}`,
```
chrisbolin: this is the key here: we were sometimes referring to fragments by their slide's index and other times by their slide's hash. Usually those two are the same. Until you set a custom slide id. 😄

Issue #512, opened 2018-04-15
Prism error in CodePane component using spectacle-scripts


sslotsky:
Title sums it up. I used `spectacle-scripts` to create the slide deck and added this codepane slide:

```javascript
        <CodePane lang="reason" source={interop} />
```

To be thorough I tried this too:

```javascript
        <CodePane lang="js" source="var foo = 'bar';" />
```

The slide crashes with the following:

```
Uncaught TypeError: Cannot read property 'buildPlaceholders' of undefined
    at prism-handlebars.js:30
    at Object.run (prism-core.js:451)
    at highlight (prism-core.js:280)
    at prism (react-live.es.js:23)
    at Editor.componentWillMount (react-live.es.js:393)
    at callComponentWillMount (react-dom.development.js:6860)
    at mountClassInstance (react-dom.development.js:6956)
    at updateClassComponent (react-dom.development.js:8325)
    at beginWork (react-dom.development.js:8966)
    at performUnitOfWork (react-dom.development.js:11798)
    at workLoop (react-dom.development.js:11827)
    at HTMLUnknownElement.callCallback (react-dom.development.js:104)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:142)
    at invokeGuardedCallback (react-dom.development.js:191)
    at replayUnitOfWork (react-dom.development.js:11302)
    at renderRoot (react-dom.development.js:11869)
    at performWorkOnRoot (react-dom.development.js:12422)
    at performWork (react-dom.development.js:12343)
    at performSyncWork (react-dom.development.js:12320)
    at requestWork (react-dom.development.js:12220)
    at scheduleWorkImpl (react-dom.development.js:12095)
    at scheduleWork (react-dom.development.js:12055)
    at scheduleRootUpdate (react-dom.development.js:12683)
    at updateContainerAtExpirationTime (react-dom.development.js:12711)
    at Object.updateContainer (react-dom.development.js:12730)
    at ReactRoot../node_modules/react-dom/cjs/react-dom.development.js.ReactRoot.render (react-dom.development.js:16030)
    at react-dom.development.js:16449
    at Object.unbatchedUpdates (react-dom.development.js:12530)
    at legacyRenderSubtreeIntoContainer (react-dom.development.js:16445)
    at Object.render (react-dom.development.js:16513)
    at Object../src/index.js (index.js:6)
    at __webpack_require__ (bootstrap 18113f8090508b41993d:678)
[...]

PR #546, opened 2018-07-11
Prism version conflict

imranolas:
Prism was nailed down to 1.6.0 due to https://github.com/FormidableLabs/spectacle/issues/512, but spectacle still depends on `react-live` which has a dependency on Prism 1.14.0

```
│ │ ├─┬ react-live@1.10.1
│ │ │ ├─┬ buble@0.19.3
│ │ │ │ ├── acorn@5.6.2 deduped
│ │ │ │ ├─┬ acorn-dynamic-import@3.0.0
│ │ │ │ │ └── acorn@5.6.2 deduped
│ │ │ │ ├─┬ acorn-jsx@4.1.1
│ │ │ │ │ └── acorn@5.6.2 deduped
│ │ │ │ ├─┬ chalk@2.4.1
│ │ │ │ │ ├── ansi-styles@3.2.1 deduped
│ │ │ │ │ ├── escape-string-regexp@1.0.5 deduped
│ │ │ │ │ └─┬ supports-color@5.4.0
│ │ │ │ │   └── has-flag@3.0.0
│ │ │ │ ├─┬ magic-string@0.22.5
│ │ │ │ │ └── vlq@0.2.3
│ │ │ │ ├── minimist@1.2.0 deduped
│ │ │ │ ├── os-homedir@1.0.2 deduped
│ │ │ │ └── vlq@1.0.0
│ │ │ ├── core-js@2.5.7 deduped
│ │ │ ├─┬ dom-iterator@1.0.0
│ │ │ │ ├── component-props@1.1.1
│ │ │ │ └── component-xor@0.0.4
│ │ │ ├─┬ prismjs@1.14.0
```

This version conflict leaves me unable to show a code slide in ReasonML. I can't help but wonder if it would be better to just load the languages in the way that the Prism team recommends https://github.com/PrismJS/prism/issues/1394

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "normalize.css": "^7.0.0",
     "prismjs": "1.6.0",
     "react-emotion": "^8.0.8",
-    "react-live": "^1.8.0-2",
+    "react-live": "^1.11.0",
     "react-redux": "^5.0.5",
     "react-transition-group": "1.2.1",
     "react-typography": "^0.16.5",
--- a/src/components/__snapshots__/code-pane.test.js.snap
+++ b/src/components/__snapshots__/code-pane.test.js.snap
@@ -106,6 +106,8 @@ exports[`<CodePane /> should render correctly. 1`] = `
                 }
               }
               onClick={undefined}
+              onCompositionEnd={undefined}
+              onCompositionStart={undefined}
               onKeyDown={undefined}
               onKeyUp={undefined}
               spellCheck="false"
--- a/src/components/__snapshots__/component-playground.test.js.snap
+++ b/src/components/__snapshots__/component-playground.test.js.snap
@@ -3,7 +3,6 @@
 exports[`<ComponentPlayground /> Should render the dark theme correctly 1`] = `
 <div
   class="react-live css-o32omv"
-  scope="[object Object]"
 >
   <div
     class="css-qtztkt"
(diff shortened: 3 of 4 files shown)


imranolas:
I've pinned prismjs to 1.6.0 in react-live and bumped the version as a stopgap for the moment.

@sslotsky I agree that a config would be annoying and I'd worry that forcing users to setup a babel plugin creates friction in what should be a relatively painless experience otherwise.


NB. Some surprising behaviour when using the hub cli - I intended to reference the issue in the PR but instead hub converted the issue _into_ a PR. I didn't even know that was a thing! Sorry everyone!

We can open a new issue to continue the discussion if preferred.

sslotsky:
@imranolas thanks! Question: willing pinning prism to 1.6.0 allow me to use ReasonML syntax? I have some concerns that whichever Prism loading code that gets executed first/last could take precedent in terms of which languages are actually loaded.

imranolas:
@sslotsky No problem. I've tested it against your minimal repro and it appears to work but I hadn't considered the issue could be loading precedence. Hopefully this is stable for now until we can find the right solution.

ryan-roemer:
@sslotsky -- It works provided _on other packages_ you add have a `prismjs` version that is different _and_ that npm flattening properly collapses to only one used `prismjs` package. If there are 2+ packages in `node_modules` tree then this temporary bandaid hack won't work.

But from `spectacle`s POV only, this is sufficient to tide us by until we get a real solution.

PR #571, opened 2018-08-17
add rel attribute into Link to prevent tabnabbing attack

wonism:
__changes__
This PR is for adding `rel="noopener noreferrer"` into `<a>` when its target is `_blank`.
Also, I changed `<Link>`'s propTypes. so that `target` can be one of `_blank`, `_self`, `_parent`, `_top`.

__references__
- [Opens External Anchors Using rel="noopener"](https://developers.google.com/web/tools/lighthouse/audits/noopener)
- [Html/Elements/a](https://www.w3.org/wiki/Html/Elements/a)
- [Reverse Tabnabbing](https://www.owasp.org/index.php/Reverse_Tabnabbing)


ebrillhart:
Thanks @wonism!

PR #576, opened 2018-08-29
Add global style documenation

badlamer:
Add missing Deck prop.


ebrillhart:
Thanks for adding, @badlamer

PR #579, opened 2018-09-05
Updating contributing guidelines

kale-stew:
#### Description

This PR introduces issue templates so contributors can open either a **Bug Report** or **Feature Request** issue. It also introduces a PR template with a Spectacle [CodeSandbox link](https://codesandbox.io/s/7wo8xv8nw0) so users can more easily demo their code.

#### Type of Change
- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -25,19 +25,15 @@ What machine are you on?
 
 ### Describe the Problem
 
-**Steps to reproduce:**
+It's easier to show us than tell us what's going wrong with your code. Because of this, we ask that you do one of three things to help us reproduce the bug:
 
-1. [first step]
-2. [second step]
-3. [you get the idea]
+1. Create a public minimal repository that we can `git clone`, with install + error reproduction steps in the README.
+2. Fork our [simple Spectacle sample Sandbox](https://codesandbox.io/s/7wo8xv8nw0), reproduce your issue in the code, and paste the link here.
+3. Open up a PR, include "WIP" and the Issue # in the title, and point us to the failing regression tests.
 
 **Expected behavior:** [What you expect to happen]
 
 **Actual behavior:** [What actually happens]
-
-**Relevant Code:**
-
-Fork our [simple Spectacle sample Sandbox](https://codesandbox.io/s/7wo8xv8nw0), reproduce your issue in the code, and paste the link here.
 
 ### Additional Information
 
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,44 @@
+<!--
+
+Have you read Formidable's Code of Conduct? By filing an Issue, you are expected to comply with it, including treating everyone with respect: https://github.com/FormidableLabs/spectacle/blob/master/CONTRIBUTING.md#contributor-covenant-code-of-conduct
+
+-->
+
+### Prerequisites
+
+**Feel free to delete this section if you have checked off all of the following.**
+
+- [ ] I've searched open [issues](https://www.github.com/FormidableLabs/spectacle/issues) to make sure I'm not opening a duplicate issue
+- [ ] This issue not specific to `spectacle-boilerplate` (those issues should be opened [here](https://github.com/FormidableLabs/spectacle-boilerplate/issues/new)).
+- [ ] I have read through the [docs](https://formidable.com/open-source/spectacle/docs/) before asking a question
+- [ ] I am using the latest version of Spectacle
+
+### Describe Your Environment
+
+What version of Spectacle are you using? (can be found by running `npm list spectacle`)
+
+What version of React are you using? (can be found by running `npm list react`)
+
+What browser are you using?
+
+What machine are you on?
+
+### Describe the Problem
+
+**Steps to reproduce:**
+
+1. [first step]
+2. [second step]
+3. [you get the idea]
+
+**Expected behavior:** [What you expect to happen]
+
+**Actual behavior:** [What actually happens]
+
+**Relevant Code:**
+
+Fork our [simple Spectacle sample Sandbox](https://codesandbox.io/s/7wo8xv8nw0), reproduce your issue in the code, and paste the link here.
+
+### Additional Information
+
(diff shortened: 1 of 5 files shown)


ebrillhart:
ACTUALLY. wait. can you add the codesandbox to the bug report too so that people can demo those more easily as well? thank youuuu @kale-stew

CONTRIBUTING.md
```diff
@@ -16,27 +16,27 @@ You will find tests for files colocated with `*.test.js` suffixes. Whenever maki
...
 
-Alternitivley install the Prettier [editor plugin](https://prettier.io/docs/en/editors.html) in your favorite editor. This is the perfered method.
+Alternatively, install the Prettier [editor plugin](https://prettier.io/docs/en/editors.html) in your favorite editor. This is the preferred method.
 
-There is also a pre-commit hook in place to lint all staged files. If any of the staged files do not conform to the eslint rules or the [prettier](https://prettier.io/) formating guidelines, your commit will fail until you resolve all outstanding issues.
+There is also a pre-commit hook in place to lint all staged files. If any of the staged files do not conform to the eslint rules or the [prettier](https://prettier.io/) formatting guidelines, your commit will fail until you resolve all outstanding issues.
 
-To resolve/fix prettier formating problems from the CLI:
+To resolve/fix prettier formatting problems from the CLI:
```
ebrillhart: so we actually have a `yarn prettier-fix` which will do the trick more easily than adjusting the individual files, we may want to update this section to basically have them run `lint` and `prettier-fix`

ebrillhart: I also for later still want to consider adding `prettier-fix` to the pre-commit hook but we've already chatted about that 🙂

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,44 @@
...
+
+What machine are you on?
+
+### Describe the Problem
+
+**Steps to reproduce:**
+
+1. [first step]
+2. [second step]
+3. [you get the idea]
```
ryan-roemer: A lot of the time, we end up asking bug submitters for:

- A public minimal repository that we can `git clone`
- With install + error reproduction steps

This saves us an enormous amount of time from somewhat manual reproductions that can have different dependencies, etc.

I propose that we _default_ to asking for that or a WIP PR with a simple failing regression test. And I think that might be a good thing as a default for all our projects.

/cc @boygirl

CONTRIBUTING.md
ryan-roemer: @ebrillhart -- I tend to vote against any pre-commit hooks, as a personal preference. The wait for any commit to succeed adds up as number of files increases and number of things fixes increases. (Maybe because I commit a lot and squash later). Fine to do whatever here, but I'd rather manually fix and have CI / a runnable check I run before final PR-ing fail me.

ebrillhart: @ryan-roemer An issue we run into a lot is people PR, CI fails, and then the submitters are unresponsive to requests to fix lint or tests or whatever and repush, which usually means duplicating the work to actually make the fix, since we don't have the same access to forked branches. I'd love to try to prevent that particular scenario, but am certainly open to alternatives to the pre-commit hook!

ebrillhart: I guess just having in the checklist that people should run the relevant scripts is probably a good compromise and we can revisit if it continues to be an issue!

.github/ISSUE_TEMPLATE/bug_report.md
ebrillhart: @ryan-roemer would you be okay with a codesandbox of the error or is that fall into the category of `somewhat manual reproduction`? If so, I'd be partial to a WIP PR - I think a separate repo could be a lot of overhead to submit an issue, especially for more jr devs

kale-stew: @ryan-roemer Do you think the CodeSandbox link might also minimize some of this bug reproduction pain? We could require at least one of these three reproduction methods is included in every bug report, so we're not limiting anyone to use of one method but we ensure there is always a way to checkout the buggy code in the issue.

Also, I realize now I had the sandbox link at the bottom of this section; it would probably be better to reword this template so it's a more prominent requirement for a bug report.

ryan-roemer: Let me back up. The _problem_ is that for a lot of repositories we get a report of a bug that looks like it may be a bug, but it isn't reproducible without a *lot* of work for the maintainers. Things that have cropped up in the past:

- Bespoke webpack config
- Dependencies variants
- Weird other libraries conflicting
- The bug not actually being a bug and something else being wrong.

For the maintainers, if something isn't immediately reproducible, that takes our valuable time trying to experiment and guess at how to set up the bug reproduction. That means trading off valuable time from everything else that we could otherwise focus on.

Contrast this with the bug opener. They're actually _hitting_ the bug. They have it! They are in the best, and most easy place to produce a working simulation (or the real repo if open source) for us of something that they can easily confirm is "the issue". On average, takes them way less time that us experimenting because they're already hitting the issue. So, the better balance if we have any resource constraints is to place a little more burden on the issue _openers_ than the maintainers so that the maintainers can field more bugs and work on fixing confirmed, reproducible bugs.

For jr devs, we most certainly can and should provide extra help with opening a failing WIP PR and/or reproduction. That's a good community investment too. But, I think we can roll in some language for that support while going for the overall goal of having "immediately reproducible" bugs that maintainers can easily verify. (It's just a goal -- won't always happen...)

So turning back to what we should ask for to meet our goal of "immediately reproducible" -- it's whatever does that! Things that should work:

- A failing regression test WIP PR
- A minimal repository with install + error reproduction instructions
- A code sandbox! (It has locked deps and is git cloneable, so it essentially **is** the above 😉 )
- ... other things!

Hope that helps!

ryan-roemer: ... and as a final thought -- not everyone has the time to do this / the capacity (for complicated stuff). And that's fine!

But, for stuff that _does_ have an immediately reproducible bug, that should influence us on what we fix first -- because it is something we can easily verify is wrong and much more readily fix. So it's not a **requirement** just a way for the community to help incentivize us to triage it higher priority / fix it sooner 😉

coopy:
These templates are great! I made a couple of suggestions. :shipit:

Issue #557, opened 2018-07-10
Slide background radial-gradient support


nalgritz:
I tried to add gradient background color to Slide tag. As `Silde` doesn't have background props so i added as follow:
```
<Slide 
  style={{ background: 'radial-gradient(ellipse at bottom, black, white)' }}
>
  { /* other components */ }
</Slide>
```
However it only shows white background color. Also tried below return transparent background but still no radial-gradient
```
<Slide 
  bgColor='transparent'
  style={{ background: 'radial-gradient(ellipse at bottom, black, white)' }} 
>
  { /* other components */ }
</Slide>
```
I wonder if it supports gradient background?

jalmarazg:
Hi! 
Actually I tried what you wanted and actually got a workaround. So, I'll first tell you why It's not working with the actual properties of the "Slide" tag:

### Reproducing the scenario
In this case I tried using ```bgColor``` as you suggested but got no results. After that, I tried creating a _theme_ with the ```createTheme``` method and tried with ```bgImage``` again with no results. 
I decided to check what could it be that cause the _radial-gradient_ not to show by inspecting the page and I got the following: 

- Using ```bgColor``` 
<img width="558" alt="im1" src="https://user-images.githubusercontent.com/9245962/43371063-f4aba6aa-934f-11e8-9a07-6d0b8e851fdf.png">

- Using ```bgImage``` 
<img width="286" alt="im2" src="https://user-images.githubusercontent.com/9245962/43371070-00004024-9350-11e8-893b-4cacd88ea1a9.png">

Checking the topic more in depth I came to realize that _radial-gradient_ only works if it's declared on css with the key _background-image_ or _background_ ( As shown in [here)](https://developer.mozilla.org/en-US/docs/Web/CSS/radial-gradient) ). After this, I tried another approach.

### Workaround
I created a CSS file with the gradient value on it:
```css
.ComponentStyle {
      background: radial-gradient(ellipse farthest-corner at 45px 45px , #00FFFF 0%, rgba(0, 0, 255, 0) 50%, #0000FF 95%);
      height: 100%;
      width: 100%;
      background-position: center;
      background-repeat: no-repeat;
      background-size: cover;
}
```
and add it as a ```className``` to the "Slide" tag like this: 
```jsx
<Slide transition={["fade"]} className = "ComponentStyle">
```

After adding the class I got the gradient background but there's some kind of margin around it. 
[...]

PR #587, opened 2018-09-17
add gbGradient prop to base.js file and base.test file

ElreyB:
#### Type of Change

Add `bgGraident` prop to base file 

### How Has This Been Tested?

Unit test on the `base.js` file and in presentation example I work with to debug

### DIFF
--- a/src/utils/base.test.js
+++ b/src/utils/base.test.js
@@ -1,7 +1,7 @@
 import { getStyles } from './base';
-/* eslint-disable */
+/* eslint-disable-next-line max-statements */
 describe('getStyles', () => {
-  let fn;
+  let generateStyles;
   let _this;
   beforeEach(() => {
     _this = {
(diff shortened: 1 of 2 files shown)


ebrillhart:
Hey @ElreyB on first glance this looks awesome - let me review more in-depth later today and I'll let you know if it's ready to merge!

Closes #557

src/utils/base.test.js
```diff
@@ -0,0 +1,221 @@
+import { getStyles } from './base';
+/* eslint-disable */
```
parkerziegler: So this will disable the entirety of ESLint for the file. We want to avoid doing this so that if future contributors alter this file they won't unintentionally be violating our ESLint setup. Prefer to always use `eslint-disable-line` or `eslint-disable-next-line` with the specific rule. In this case, it looks like you're violating the `max-statements` rule, which is totally fine for a test file. So just replace `/* eslint-disable */` with `eslint-disable-next-line max-statements` here.

src/utils/base.test.js
```diff
@@ -0,0 +1,221 @@
...
+    });
+
+    it('should assign bgSize value to backgroundSize', () => {
+      const bgSizeValue = 'auto';
+      _this.props.bgSize = bgSizeValue;
+      const res = fn();
+      expect(res.backgroundSize).toEqual(bgSizeValue);
+    });
+
+    it("should assign 'cover' to backgroundSize", () => {
```
parkerziegler: You'll get an ESLint `quotes` violation here and on L190, so just disable it line by line, i.e. `eslint-disable-next-line quotes`.

src/utils/base.test.js
```diff
@@ -0,0 +1,221 @@
+import { getStyles } from './base';
+/* eslint-disable */
+describe('getStyles', () => {
+  let fn;
```
parkerziegler: I know I was the one who demo'ed using `fn` and `res` 🤦‍♂️ , but let's use variable names more consistent with what the function actually does. `res` in particular has significant meaning on the server-side. What do you think about this:

~~`fn`~~ -> `generateStyles`
~~`res`~~ -> `styles`

src/utils/base.test.js
```diff
@@ -0,0 +1,221 @@
...
+
+  describe('fontFamily', () => {
+    it('should assign fontFamily to default font Helvetica', () => {
+      _this.props.textFont = _this.context.styles.fonts.primary;
+      const res = fn();
+      expect(res.fontFamily).toEqual('Helvetica');
+    });
+  });
+
+  // Need to figure out how to test for textSize
```
parkerziegler: I find it helpful to prefix comments like this with `@TODO`. For example:

```
// @TODO add test for textSize
```

Makes comments like this easily searchable across the codebase!

src/utils/base.test.js
```diff
@@ -0,0 +1,221 @@
...
+  describe('height', () => {
+    it('should assign height value to height', () => {
+      const heightValue = 'scroll';
+      _this.props.height = heightValue;
+      const res = fn();
+      expect(res.height).toEqual(heightValue);
+    });
+  });
+});
+/* eslint-enable */
```
parkerziegler: No need to re-enable here since there's no more to this file – see comments about selective `disabling` above!

parkerziegler:
Looks awesome! @ebrillhart I think this looks good to merge, but would prefer you give the final approval!

Issue #540, opened 2018-06-05
componentWillReceiveProps is deprecated since React 16.3.0


ryan-roemer:
We've disabled these, but:

```
/Users/rye/scm/fmd/spectacle/src/components/anim.js
  11:1  error  componentWillReceiveProps is deprecated since React 16.3.0, use UNSAFE_componentWillReceiveProps instead, see https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops  react/no-deprecated
```

ebrillhart:
closed with #591

PR #591, opened 2018-09-18
[WIP] Remove deprecated lifecycles

ElreyB:
Fixes # (issue) #540 

#### Type of Change

- [x] Remove deprecated componentWillReceiveProps with getDerivedStateFromProps
- [x] Remove deprecated componentWillMount with componentDidMount

- [x] My code follows the style guidelines of this project (I have run `yarn prettier-fix && yarn lint`)
- [x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/src/components/component-playground.js
+++ b/src/components/component-playground.js
@@ -139,10 +139,6 @@ function getEnhancedScope(scope = {}) {
   return { Component, ...scope };
 }
 
-// TODO(540): Refactor to non-deprecated lifecycle methods.
-// https://github.com/FormidableLabs/spectacle/issues/540
-// - componentWillReceiveProps
-// eslint-disable-next-line react/no-deprecated
 class ComponentPlayground extends Component {
   constructor() {
     super(...arguments);
@@ -157,13 +153,13 @@ class ComponentPlayground extends Component {
     };
   }
 
-  static getDerivedStateFromProps(nextProps, preState) {
+  static getDerivedStateFromProps(nextProps, prevState) {
     const updatedState = {};
-    if (nextProps.code !== preState.code) {
+    if (nextProps.code !== prevState.code) {
       const code = (nextProps.code || defaultCode).trim();
       updatedState.code = code;
     }
-    if (nextProps.scope !== preState.scope) {
+    if (nextProps.scope !== prevState.scope) {
       const scope = getEnhancedScope(nextProps.scope);
       updatedState.scope = scope;
     }
@@ -176,14 +172,14 @@ class ComponentPlayground extends Component {
   }
 
   componentDidUpdate() {
-    this.playgroundsetState();
+    this.playgroundSetState();
   }
 
   componentWillUnmount() {
     window.removeEventListener('storage', this.syncCode);
   }
 
-  playgroundsetState() {
+  playgroundSetState() {
     if (this.props.code) {
       const code = (this.props.code || defaultCode).trim();
       this.setState({ code });
--- a/src/components/component-playground.js
+++ b/src/components/component-playground.js
@@ -1,4 +1,5 @@
 import React, { Component } from 'react';
+import isEmpty from 'lodash/isEmpty';
 import PropTypes from 'prop-types';
 import styled, { css } from 'react-emotion';
 import { defaultCode } from '../utils/playground.default-code';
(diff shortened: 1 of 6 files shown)


ebrillhart:
This actually should, ideally, _not_ be a breaking change in the sense that it should not affect the functionality for any users of the library, only how the code is written and runs under the hood!

ElreyB:
@ebrillhart cool beans. I think I just pick the wrong option for that section.

src/components/manager.js
```diff
@@ -57,6 +57,44 @@ const StyledTransition = styled(ReactTransitionGroup)({
...
+          reference.goTo = child.props.goTo;
+        }
+        slideReference.push(reference);
+      });
+    }
+  });
+  return slideReference;
+}
+
 // TODO(540): Refactor to non-deprecated lifecycle methods.
```
kale-stew: Could you please remove this `//@TODO` block? 🙂

src/components/manager.js
```diff
@@ -57,6 +57,44 @@ const StyledTransition = styled(ReactTransitionGroup)({
   transformStyle: 'flat'
 });
 
+function buildSlideReference(props) {
```
kale-stew: Could I ask why you opted to move this method outside the component?

src/components/heading.js
```diff
@@ -69,9 +69,17 @@ export default class Heading extends Component {
     window.addEventListener('load', this.resize);
     window.addEventListener('resize', this.resize);
   }
-  componentWillReceiveProps() {
-    this.resize();
+
+  static getDerivedStateFromProps(nextProps, prevState) {
+    return nextProps.fit !== prevState.fit ? { fit: nextProps.fit } : null;
```
kale-stew: This looks great!

src/components/component-playground.js
```diff
@@ -156,26 +157,42 @@ class ComponentPlayground extends Component {
...
-    }
+  componentDidUpdate() {
+    this.playgroundsetState();
   }
 
   componentWillUnmount() {
     window.removeEventListener('storage', this.syncCode);
   }
 
+  playgroundsetState() {
```
kale-stew: Small nit: could you update this to proper camelCase? I think `playgroundSetState` is just a bit more pronounced

src/components/component-playground.js
```diff
@@ -156,26 +157,42 @@ class ComponentPlayground extends Component {
     };
   }
 
+  static getDerivedStateFromProps(nextProps, preState) {
```
kale-stew: Another small nit: I think `prevState` is more conventional (you can see the same pattern [here](https://github.com/FormidableLabs/spectacle/pull/591/files#diff-16ef0e6c92055f5ab24dd4cd04a48742R77))

src/components/manager.js
ElreyB: @kale-stew [getDerviedStateFromProps](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) does not have access to the component instance. Since `buildSlideReference` didn't have any component instances in the method itself, I opted to take it out of the class and use it as a regular method that the class can use.

src/components/text.js
```diff
@@ -54,8 +50,14 @@ export default class Text extends Component {
     window.addEventListener('load', this.resize);
     window.addEventListener('resize', this.resize);
   }
-  componentWillReceiveProps() {
-    this.resize();
+  static getDerivedStateFromProps(nextProps, prevState) {
+    return nextProps.fit !== prevState.fit ? { fit: nextProps.fit } : null;
+  }
+
+  componentDidUpdate(prevState, prevProps) {
```
parkerziegler: This function signature is incorrect – `componentDidUpdate` takes `prevProps` as the first argument and `prevState` as an optional second argument. Unfortunately, this means that the `prevProps` reference below is actually to `prevState` 😱.

src/components/heading.js
```diff
@@ -70,7 +70,7 @@ export default class Heading extends Component {
     return nextProps.fit !== prevState.fit ? { fit: nextProps.fit } : null;
   }
 
-  componentDidUpdate(prevState, prevProps) {
+  componentDidUpdate(prevProps) {
     if (prevProps.fit !== this.props.fit) {
```
parkerziegler: One thing to note here. By removing that first argument, you actually changed what `prevProps` means in the method. In the first iteration:

```
componentDidUpdate(prevState, prevProps) {
   if (prevProps.fit !== this.props.fit) {
      //
   }
}

`prevProps`, because of its position as the second argument, actually refers to `prevState`.

In the new iteration:

```
componentDidUpdate(prevProps) {
   if (prevProps.fit !== this.props.fit) {
      //
   }
}
```

`prevProps` correctly refers to `prevProps`, but your function hasn't changed at all to reflect that update. I think that's likely fine based on the other patterns you've implemented here, but want to make sure you understand the difference :)

parkerziegler:
@ebrillhart @kale-stew I symlinked this locally against one of my Spectacle presentations and it worked as expected. I also ran the demo locally after fresh install and all seemed normal there. I think we are likely good for a merge.

It may be worth noting in the CHANGELOG that users will need to ensure they have React and ReactDOM ^16.3.1 as their local dependency (earliest version with stable `getDerivedStateFromProps`). Not a huge issue, people should hopefully get that update automatically if they `yarn upgrade` / `npm update` or start fresh from the boilerplate, but I can see folks just upgrading their `spectacle` dep, running `yarn` / `npm i`, and not getting the React / ReactDOM update b/c of `yarn.lock` / `package-lock.json`.

ebrillhart:
I will merge this and make a note of dependency stuff in the release notes for next release!

PR #611, opened 2018-11-06
Remove specialized zoom handling and margins for slide content

ebrillhart:
This removes custom zoom handling and reverts to using default browser zoom for more predictable behavior and better accessibility. Can discuss later if this change causes lots of issues for people.

Also uses a standard single margin for slide content instead of calculating based on margin/zoom level, as that was creating weird zoom issues and often doubling the amount of padding/margin.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
     <meta charset="UTF-8">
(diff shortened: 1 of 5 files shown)


justsml:
Tested, looks great! Thanks @ebrillhart

PR #624, opened 2018-12-21
Remove warnings from test suite

rawrmonstar:
Removes some of the warnings that occurred running the test suite and runs `prettier-fix`. No breaking changes.


ebrillhart:
approved once CI passes!

Issue #639, opened 2019-01-09
Update eslint ecosystem


ebrillhart:
update the following to current:
```
"eslint": "^4.19.0",
    "eslint-config-formidable": "^3.0.0",
    "eslint-config-prettier": "^2.9.0",
    "eslint-plugin-filenames": "^1.2.0",
    "eslint-plugin-import": "^2.9.0",
    "eslint-plugin-react": "^7.7.0",
```

Issue #638, opened 2019-01-09
Update react-redux


ebrillhart:
Upgrade to `react-redux` v6+ and handle breaking changes

Issue #635, opened 2019-01-09
Upgrade to react-transition-group v2


ebrillhart:
### Description

Make migration to `react-transition-group v2+

### Links / References

[Migration Docs](https://github.com/reactjs/react-transition-group/blob/HEAD/Migration.md)

Issue #634, opened 2019-01-09
Upgrade Emotion to v.10+ (w/ related switch from react-emotion)


ebrillhart:
### Description

Upgrade to `emotion` v.10+ and switch from `react-emotion` to `@emotion/core` and `@emotion/styled` where needed.

Will also require an update to `babel-plugin-emotion`

PR #643, opened 2019-01-11
Update Packages

### LINKED ISSUE #631: Feature: Upgrade babel libraries to non-beta.
### Description

We've got a lot of `-beta` babel deps in `package.json` like:

```js
"@babel/core": "^7.0.0-beta.42",
```

This causes our users to sometimes hit things like #522 which can be tricky to resolve. Also, we're on beta when all these babel libs are on `7.2.2`

### Proposal

Upgrade all `@babel` libs to current (presently `7.2.2`) and any related babel-supporting libraries to current.

### Links / References

- https://github.com/FormidableLabs/spectacle/issues/522

ebrillhart:
### Description

A lot of our deps were pretty out of date, so this updates (almost) all of them. It should fix #631, specifically, and also get us in a better place moving forward for more regular maintenance of the package.json.

Created some follow-up issues for updates w/ breaking changes:
- #634 
- #635 
- #638 
- #639 
- #648 

Fixes #631


package.json
```diff
@@ -46,77 +46,77 @@
...
-    "history": "^4.6.1",
-    "lodash": "^4.17.4",
-    "marksy": "^6.0.2",
-    "normalize.css": "^7.0.0",
-    "prismjs": "1.6.0",
+    "history": "^4.7.2",
+    "lodash": "^4.17.11",
+    "marksy": "^6.1.0",
+    "normalize.css": "^8.0.1",
+    "prismjs": "1.15.0",
```
ryan-roemer: **WARNING**: `prismjs` has to match whatever is in `react-live` **exactly** or horrible things happen. (Also, if any other transitive dep brings `prismjs` in that _also_ has to match).

See, e.g. #546 #558 #516

package.json
```diff
@@ -46,77 +46,77 @@
...
     "to-style": "^1.3.3",
-    "victory-core": "^15.2.0"
+    "victory-core": "^31.1.0"
   },
   "peerDependencies": {
-    "prop-types": "^15.6.0",
-    "react": "^16.0.0",
-    "react-dom": "^16.0.0"
+    "prop-types": "^15.6.2",
+    "react": "^16.7.0",
```
ryan-roemer: Do we want to require 16.7 above? Or can we support anything below that? Ideally all `peerDependencies` are as permissive as possible to allow folks to use this package that aren't at the latest version if everything is still correct.

Same for all other deps here.

package.json
```diff
@@ -46,77 +46,77 @@
...
-    "@babel/plugin-proposal-class-properties": "^7.0.0-beta.42",
-    "@babel/plugin-proposal-object-rest-spread": "^7.0.0-beta.42",
-    "@babel/polyfill": "^7.0.0-beta.42",
-    "@babel/preset-env": "^7.0.0-beta.42",
-    "@babel/preset-react": "^7.0.0-beta.42",
-    "babel-core": "^7.0.0-bridge.0",
-    "babel-eslint": "^8.2.2",
-    "babel-jest": "^22.1.0",
-    "babel-loader": "^8.0.0-beta.2",
+    "@babel/cli": "^7.2.3",
```
ryan-roemer: Love it! Finally an end to all the related `-beta` issues!

package.json
```diff
@@ -46,77 +46,77 @@
...
+    "@babel/core": "^7.2.2",
+    "@babel/plugin-proposal-class-properties": "^7.2.3",
+    "@babel/plugin-proposal-object-rest-spread": "^7.2.0",
+    "@babel/polyfill": "^7.2.5",
+    "@babel/preset-env": "^7.2.3",
+    "@babel/preset-react": "^7.0.0",
+    "babel-core": "^7.0.0-0",
```
ryan-roemer: This should be removed. `@babel/core` is the replacement after 7 when the `@babel` namespace replaced `babel-` in all the official Babel project packages.

package.json
```diff
@@ -46,77 +46,77 @@
...
-    "webpack-dev-server": "^3.1.1"
+    "url-loader": "^1.1.2",
+    "webpack": "^4.28.4",
+    "webpack-cli": "^3.2.1",
+    "webpack-dev-server": "^3.1.14"
   },
   "resolutions": {
-    "babel-core": "7.0.0-bridge.0",
-    "source-map": "0.6.1"
+    "babel-core": "^7.0.0-0",
```
ryan-roemer: Should likely be removed too...

package.json
ebrillhart: okay - will make the latest major version the base-level requirement for peer deps, which I think is reasonable!

package.json
ebrillhart: The latest version of babel-jest unfortunately still requires this bridge version as a peer dependency, so we'll need to leave in for now. Otherwise, all tests fail.

ebrillhart: huh, never mind, this must have been an extension of issues related to updating source-map. removed!

package.json
```diff
@@ -46,76 +46,75 @@
...
-    "webpack-cli": "^3.1.2",
-    "webpack-dev-server": "^3.1.1"
+    "url-loader": "^1.1.2",
+    "webpack": "^4.28.4",
+    "webpack-cli": "^3.2.1",
+    "webpack-dev-server": "^3.1.14"
   },
   "resolutions": {
-    "babel-core": "7.0.0-bridge.0",
+    "babel-core": "^7.0.0-0",
```
ebrillhart: It does need this as a resolution for the babel-jest dependency, so I'm leaving it here until that changes

PR #649, opened 2019-01-25
Fixes no slide separation with Windows line endings

debel:
### Description

When using markdown files saved on windows as slides everything appeared as a single slide. This patch fixes it so '.md' files with both UNIX and Windows line endings are rendered properly.

#### Type of Change

Bug fix (non-breaking change which fixes an issue)

### How Has This Been Tested?

Tested locally with an `.md` file saved with Windows line endings

### DIFF
--- a/src/components/markdown-slides.js
+++ b/src/components/markdown-slides.js
@@ -3,7 +3,7 @@ import Slide from './slide';
 import Markdown from './markdown';
 
 const transformStringsIntoJSX = function(strings) {
-  return strings.split(/\n---\n/).map((markdown, index) => (
+  return strings.split(/\r?\n---\r?\n/).map((markdown, index) => (
     <Slide key={`md-slide-${index}`}>
       <Markdown>{markdown}</Markdown>
     </Slide>


src/components/markdown-slides.js
```diff
@@ -3,7 +3,7 @@ import Slide from './slide';
 import Markdown from './markdown';
 
 const transformStringsIntoJSX = function(strings) {
-  return strings.split(/\n---\n/).map((markdown, index) => (
+  return strings.split(/\r?\n---\r?\n/).map((markdown, index) => (
```
kitten: I'd propose the newlines to be defined as `(\r\n|\r|\n)` which is more correct, since carriage returns `\r` and line feeds can both be used individually. This would cover all styles of line endings.

Do we also want to introduce `---+` to cover the possibility of separators being longer than necessary?

cc @ebrillhart

debel: @kitten

AFAIK, no modern system uses `\r` only, but I can update the PR if needed 🙂

ebrillhart: Phil knows more about the specific carriage return stuff than I do, so I'll defer to him on that particular issue.

With regards to the longer than necessary separators, I do think it would be a good idea to introduce catches - @debel if you'd like to work that into this PR then feel free, otherwise I'll open a separate issue.

ebrillhart:
@debel let me know when you feel that this PR is feature complete and I'll do a final review and merge!

kale-stew:
@debel bumping Emma's request above ☝️ I am taking over maintenance of Spectacle and would love to get this moving if you are ok with it.

kale-stew:
Closing as outdated, but we will keep a look out for markdown slide bugs as we release [v6](https://github.com/FormidableLabs/spectacle/issues/716). 👀

PR #652, opened 2019-02-15
Custom listItem enumerators with unicode

### LINKED ISSUE #437: [Feature Request]: custom ListItem enumerators
Hello! I wanted to have custom bullets on my listitems, eg having an X or a ✅. so ideally it would look something like this

```javascript
              <List type="X">
                <ListItem>Know every LOC</ListItem>
                <ListItem>Flow</ListItem>
                <ListItem>advanced Git</ListItem>
                <ListItem>CS Degree</ListItem>
              </List>
```

johannam999:
### Description

This feature allows adding custom bullets to `List` and `ListItem` elements. If the `bulletStyle` prop is passed to `List` component it changes the default enumerators styling for all `List` elements unless `bulletStyle` prop for `ListItem` is passed at the same time, which overwrites the styling of the parent element for particular `ListItem` only.
The user can either pass any Unicode number or use a few pre-defined strings to execute the new styling.

Fixes #437

#### Type of Change

### How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] My code follows the style guidelines of this project (I have run `yarn prettier-fix && yarn lint`)
- [x] I have made corresponding changes to the documentation

### DIFF
--- a/src/components/list-item.js
+++ b/src/components/list-item.js
@@ -3,7 +3,30 @@ import PropTypes from 'prop-types';
 import { getStyles } from '../utils/base';
 import styled from 'react-emotion';
 
-const StyledListItem = styled.li(props => props.styles);
+const bulletStyles = {
+  star: '\\2605',
+  classicCheck: '\\2713',
+  greenCheck: '\\2705',
+  arrow: '\\27a4',
+  cross: '\\274C'
+};
+const getListStyle = props => {
+  if (props.bulletStyle) {
+    return [
+      { listStyleType: 'none' },
+      `&::before {
+        content: '${bulletStyles[props.bulletStyle]}';
+        margin: 0 25px;
+    }`
+    ];
+  }
+  return undefined;
+};
+
+const StyledListItem = styled.li(
+  props => props.styles,
+  props => getListStyle(props)
+);
 
 export default class ListItem extends Component {
   render() {
@@ -11,6 +34,7 @@ export default class ListItem extends Component {
     return (
       <StyledListItem
         className={this.props.className}
+        bulletStyle={this.props.bulletStyle}
         styles={[
           this.context.styles.components.listItem,
           getStyles.call(this),
@@ -25,6 +49,7 @@ export default class ListItem extends Component {
 }
 
 ListItem.propTypes = {
+  bulletStyle: PropTypes.string,
   children: PropTypes.node,
   className: PropTypes.string,
   style: PropTypes.object
--- a/src/components/list-item.js
+++ b/src/components/list-item.js
@@ -12,7 +12,6 @@ export default class ListItem extends Component {
     return (
       <StyledListItem
         className={this.props.className}
-        bulletStyle={this.props.bulletStyle}
         styles={[
           this.context.styles.components.listItem,
           getStyles.call(this),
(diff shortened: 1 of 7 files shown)


src/components/list.js
```diff
@@ -3,6 +3,51 @@ import PropTypes from 'prop-types';
...
+// Is it okay to export from here?  Should they go in different files?
+export const bulletStyles = {
+  star: '\\2605',
+  classicCheck: '\\2713',
+  greenCheck: '\\2705',
+  arrow: '\\219d',
+  cross: '\\274C'
+};
+
+export const getBulletStyle = (bulletStyle, isListItemStyle) => {
```
johannam999: @ebrillhart I believe this whole piece of styling should go to the body.js file, however, I was not able to move it with Mariano's help as this file required `getBulletStyle ` function to return an object which is not an option at this moment as we are returning an object and a string at once. Is there any built in method I can use to convert it to an object or it is ok to keep it here for now?

src/components/list.js
```diff
@@ -3,6 +3,51 @@ import PropTypes from 'prop-types';
...
+  if (bulletStyle) {
+    const content = bulletStyles[bulletStyle] || `\\${bulletStyle}`;
+
+    if (isListItemStyle === true) {
+      return [
+        { listStyleType: 'none' },
+        `&::before {
+          content: '${content}' !important;
```
johannam999: @ebrillhart I had to use `!important` as List element had higher specificity than ListItem and I wanted to overwrite List styling with ListItem styling in some cases. Mariano mentioned that in this scenario it will be ok to use it. Let me know what you think.

src/components/list-item.js
```diff
@@ -11,11 +12,13 @@ export default class ListItem extends Component {
...
       <StyledListItem
         className={this.props.className}
+        bulletStyle={this.props.bulletStyle}
         styles={[
           this.context.styles.components.listItem,
           getStyles.call(this),
           typefaceStyle,
-          this.props.style
+          this.props.style,
+          getBulletStyle(this.props.bulletStyle, true)
```
kale-stew: Adding a check for the prop `bulletStyle` here might solve some of the snapshot issues where an empty array is being passed into the `styles` obj

src/components/list.js
```diff
@@ -3,6 +3,50 @@ import PropTypes from 'prop-types';
...
+          width: 20px;
+          font-size: 20px;
+          text-align: center;
+          vertical-align: middle;
+        }`
+      ];
+    }
+  }
+
+  return [];
```
kale-stew: Although, you could just opt to not pass an empty array if none of the above styles are applicable and instead `return;`

src/components/list.js
```diff
@@ -3,6 +3,50 @@ import PropTypes from 'prop-types';
...
+      return [
+        { listStyleType: 'none' },
+        `li::before {
+          content: '${content}';
+          display: inline-block;
+          margin-right: 40px;
+          width: 20px;
+          font-size: 20px;
+          text-align: center;
+          vertical-align: middle;
```
mariano-formidable: Hmmm... it looks like these styles are shared by both the `li::before` and `&::before` style blocks:

```
display: inline-block;
margin-right: 40px;
width: 20px;
font-size: 20px;
text-align: center;
vertical-align: middle;
```

I feel there should be a way to share them so we have don't two separate style blocks to maintain should we need to update them styles 😎

src/components/list.js
ebrillhart: we can keep it like this for now!

Issue #457, opened 2018-01-21
Slide maxWidth seems to not work


MikeDabrowski:
Great job with spectacle. Really appreciate it!

I got a problem with long title words not being centered. They just overflow container and disappear to the right. I tried to set maxWidth on Slide tag as said in the docs but it does nothing. I had to change contentWidth of Deck to workaround this issue. 

What is the correct to handle long title words? 1000px is ok but text should stay centered after wrap no matter what.

![image](https://user-images.githubusercontent.com/17836911/35198584-3e91bf0a-fef1-11e7-821b-2ecaa1c54e13.png)

samwhale:
Hi @MikeDabrowski, could you give me a small pen that I could take a look at to reproduce the problem?

ebrillhart:
Closed - maxWidth is no longer a prop, see #667 for suggestion of what to use instead

PR #667, opened 2019-04-02
update README to remove maxWidth props

ashika01:
### Description

Remove `maxWidth`, `maxHeight` prop from README.md. This is going to close issue #457. Removing the props since it is not being used anywhere in the codebase. 

It makes more sense to control the height and width using `contentHeight` and `contentWidth` (as single source) rather than introducing this new props.

Fixes # 457

#### Type of Change

-  This change requires a documentation update


ashika01:
This should ideally close the PR #554 and issue #457

ebrillhart:
closes #457

PR #686, opened 2019-06-10
Update deck.js

KenzoIkeizume:
### Description

Including a prop types on deck.

Fixes # (issue)

#### Type of Change

### How Has This Been Tested?

Pass a type in a deck component and view if breaks.


kale-stew:
Thank you for updating these definitions!

I don't believe `contentWidth` or `contentHeight` is

1) set on the Deck, instead it's a propType of the Slide
2) is meant to be modified, it's more of an internal tracker for the component to render a magic-wrapper, fullscreen, or a mobile view.

If you remove these two types from the Deck, I would be happy to merge

src/components/deck.js
```diff
@@ -28,7 +28,10 @@ export default class Deck extends Component {
     autoplayLoop: PropTypes.bool,
     autoplayOnStart: PropTypes.bool,
     children: PropTypes.node,
+    contentHeight: PropTypes.number,
```
kale-stew:
```suggestion

```

src/components/deck.js
```diff
@@ -28,7 +28,10 @@ export default class Deck extends Component {
...
+    contentWidth: PropTypes.number,
```
kale-stew:
```suggestion
```

KenzoIkeizume:
I added the suggestion, but I committed another. Could you view if this can be applied? Please :)

kale-stew:
I do have to say, I'm not a fan of renaming `props` to `properties` in a single file, as it doesn't mesh with our conventions throughout the rest of the codebase, but it doesn't break anything! So to avoid more back and forth on this, I'm going to just merge. 👍

Issue #55, opened 2015-08-20
Doesn't work in FF: CSS2Properties doesn't have an indexed property setter in FF


suevalov:
The bug is reproducible since 0.0.3. The problem is only during Deck mode. Presenter mode works just fine. This breaks JS on the page and controls don't work at all.

For 0.0.2 everything works fine.

kenwheeler:
What isn't working? I just fired up my FF with master and didn't see a problem.

suevalov:
Sorry for insufficient explanation. 

Transition between slides doesn't work. Basically, slide is rendered, but with these errors. After this you can't translate between slides.
I reproduced the bug on FF 35.0.1 on MacOS, but today I was notified by my colleague about the same problem on latest version on Windows. My presentation just didn't work in his FF.

Just jumped to the latest revision and launched demo presentation. 
![mozilla_firefox_and_doesn_t_work_in_ff__css2properties_doesn_t_have_an_indexed_property_setter_in_ff_ _issue__55_ _formidablelabs_spectacle](https://cloud.githubusercontent.com/assets/3672221/9395877/c3bfa500-4799-11e5-8710-04147b38fd5d.png)

kenwheeler:
What code is it erroring on?

kenwheeler:
Whats line 160?

suevalov:
It goes to 1385 line, which is CSSPropertyOperations React module.

The only thing I could find is https://github.com/anselmh/object-fit/issues/5, but i didn't help to localize the problem in Spectacle.

suevalov:
Did you manage to reproduce the issue?

kenwheeler:
I have not been able to reproduce in FF. Do you know the line in Spectacle that this originates from?

kenwheeler:
Any news on this? Closing, will reopen pending more details.

elbaro:
Ran into this old issue. Download the example [one-page.html](https://github.com/FormidableLabs/spectacle/blob/master/one-page.html) and open it with FF. The page 13 is broken.

VirtualDOMinic:
Closing this issue as I only reopened it to address @elbaro 's more recent comment!

One pager should work as expected in latest master (thanks to #687 )

PR #687, opened 2019-06-11
Update one-page.html to use spectacle 5

VirtualDOMinic:
### Description

- Fixes CSSStyleDeclaration index property setter error.
- Runs prettier on `one-page.html`, too.

Loosely relates to issue #55 - primarily the last comment on there (from @ebello on 17 Apr)

I feel the original bug in that issue has likely been resolved by the latest version (not directly confirmed), but the issue raised in the comment is fixed by this.

#### Type of Change

Bug fix (non-breaking change which fixes an issue)

### How Has This Been Tested?

Tried `one-page.html` before and after: 

- Before, the error occurs on `one-page.html#/12` on Chrome 74\* & Firefox 67\*\*
  - Error doesn't occur on Safari
- After, the error doesn't occur on any slide on Chrome, Firefox, Safari.
  - Works fine with both the `one-page.js` from versions `^4` and `^5` as they are identical, but have updated this to version 5 for consistency

\* Error: `TypeError: Failed to set an indexed property on 'CSSStyleDeclaration': Index property setter is not supported.`
\*\* Error: `TypeError: "CSS2Properties doesn't have an indexed property setter for '0'"`

Checklist: complete

### DIFF
--- a/one-page.html
+++ b/one-page.html
@@ -29,276 +29,276 @@
     <script src="https://unpkg.com/spectacle@^5/dist/spectacle.js"></script>
     <script src="https://unpkg.com/spectacle@^5/lib/one-page.js"></script>
     <script type="text/spectacle">
-            () => {
-              // --------------------------------------------------------------------
-              // Code stuff to use in the presentation.
-              // (We can't link in / require files, but we can approximate here.)
-              // --------------------------------------------------------------------
-              // Customize the default theme.
-              const { themes: { defaultTheme } } = Spectacle;
-              const theme = defaultTheme({
-                primary: "#ff4081"
-              });
+      () => {
+        // --------------------------------------------------------------------
+        // Code stuff to use in the presentation.
+        // (We can't link in / require files, but we can approximate here.)
+        // --------------------------------------------------------------------
+        // Customize the default theme.
+        const { themes: { defaultTheme } } = Spectacle;
+        const theme = defaultTheme({
+          primary: "#ff4081"
+        });
 
-              // Example source code for a source code slide.
-              const exampleDeck = `
-      return (
-        <Deck transition={['zoom','slide']} transitionDuration={800}>
-          <Slide bgColor="primary">
-            <Heading size={1} fit caps>
-              React Presentations
-            </Heading>
-            <Heading size={2} fit caps>
-              Written In React
-            </Heading>
-          </Slide>
-          <Slide bgColor="black">
-            <Heading size={1} fit textColor="primary" textFont="secondary">
-              Wait What?
-            </Heading>
-          </Slide>
-          <Slide bgColor="primary" textColor="black" align="center top">
-            <Heading size={1} textColor="black" textFont="primary">
-              Thats right
-            </Heading>
-            <List>
-              <ListItem>Inline style based theme system</ListItem>
-              <ListItem>Autofit Text</ListItem>
-              <ListItem>PDF Export</ListItem>
-            </List>
-          </Slide>
-        </Deck>
-      )
-              `.trim();
+        // Example source code for a source code slide.
+        const exampleDeck = `
+return (
+  <Deck transition={['zoom','slide']} transitionDuration={800}>
+    <Slide bgColor="primary">
+      <Heading size={1} fit caps>
+        React Presentations
+      </Heading>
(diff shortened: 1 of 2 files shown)


one-page.html
```diff
@@ -1,292 +1,304 @@
...
+                      <Cite>Ken Wheeler</Cite>
+                    </BlockQuote>
+                  </Slide>
+                  <Slide transition={["spin", "zoom"]} bgColor="tertiary">
+                    <Heading caps fit size={1} textColor="primary">
+                      Inline Markdown
+                    </Heading>
+                    <Markdown>
+                      {`
+      ![Markdown Logo](https://cdn.rawgit.com/FormidableLabs/spectacle/master/example/assets/markdown.png)
```
ryan-roemer: GLOBAL: The markdown stuff cannot be indented or it won't render. It either has to be deindented like before or use a dedent function something like the following that preserves relative indentation: 

```js
        // Indent to smallest non-empty whitespace level and trim start / end of string.
        const indent = val => {
          let prefix = null;
          return (val || "").split("\n").filter(line => {
            const [cur, remainder] = (line.match(/^([ ]*)([^ ]+)/) || []).slice(1);
            return remainder ? (prefix = null === prefix || cur.length < prefix.length ? cur : prefix,
            !0) : null !== prefix;
          }).map(line => prefix ? line.replace(prefix, "") : line).join("\n").trimRight();
        };
```

VirtualDOMinic: Thanks for spotting this! Weirdly whitespace-only diffs didn't show up in my vs code version control tab... and then I glossed over the different look of those markdown slides 🤦‍♂ .

Sorry for a potential over-simplification of the ustility function, but would "sanitising" `Markdown` and `MarkdownSlides` with `.replace(/^[ \t]+/gm, '')` not be sufficient? It seems to work for both of these components in this use case.

If so, would you prefer I add this sanitisation to the components themselves, or try to implement a prettier ignore for these sections instead? IMO the former would be more robust

VirtualDOMinic: For clarity, my usage of that `replace()` is in the two spots contained below:

```js 
<Markdown>
                      {`
      ![Markdown Logo](https://cdn.rawgit.com/FormidableLabs/spectacle/master/example/assets/markdown.png)

      You can write inline images, [Markdown Links](http://commonmark.org), paragraph text and most other markdown syntax
      * Lists too!
      * With ~~strikethrough~~ and _italic_
      * And let's not forget **bold**
                      `.replace(/^[ \t]+/gm, '')}
                    </Markdown>
                  </Slide>
                  {
                    MarkdownSlides(`
      #### Create Multiple Slides in Markdown
      All the same tags and elements supported in <Markdown /> are supported in MarkdownSlides.
      ---
      Slides are separated with **three dashes** and can be used _anywhere_ in the deck. The markdown can either be:
      * A Tagged Template Literal
      * Imported Markdown from another file
                    `.replace(/^[ \t]+/gm, ''))
                  }
```

ryan-roemer: Yep. That's why code snippet has to do the `prefix` inference and uses the heuristic of "first non-whitespace level of indentation is the indentation to strip of the rest".

Put more concisely, we need to transform:

```md
    * one
    * two
        * a
        * b
```

to this:

```md
* one
* two
    * a
    * b
```

not this:

```md
* one
* two
* a
* b
```

I'm not sure if it belongs in spectacle core or just as a snippet in `one-page.html` itself, but I think it's either that or _don't_ autoformat whitespace which in our case creates a correctness problem since the spacing is meaningful.

That snippet I pasted is pretty battle-honed as I use it in nearly all my one page presentations...

ryan-roemer: That said, if we're doing it officially in spectacle (probably don't need for one-page.html because that's up to the user, we should add specific unit tests to cover corner cases like:

- Indentation token to strip occurs again within text and needs to be preserved
- Windows line breaks
- Tabs instead of spaces
- ...etc.

I'm cool with whatever we decide...

VirtualDOMinic: Saving this discussion in my bookmarks to take into account during re-write! 

For now, have just undone all the prettier indentation changes inside the `<script>`

VirtualDOMinic:
@ryan-roemer Travis now passing. Let me know if you're happy for this to merge!

PR #689, opened 2019-06-12
Rewrite: add webpack mdx

VirtualDOMinic:
First attempt at adding MDX support to rewrite.

Should we add `es/` and `dist/` to `.gitignore`?


ryan-roemer:
Ignore es dist and lib. Only src should be in source

ryan-roemer:
And they’re all git ignored correctly already. I’m not sure how they could get added in a branch?

ryan-roemer:
Ah @kale-stew looks like task/rewrite branch is missing master's https://github.com/FormidableLabs/spectacle/blob/master/.gitignore Suggest we get that going in there sooner than later ;)

VirtualDOMinic:
> And they’re all git ignored correctly already. I’m not sure how they could get added in a branch?

@ryan-roemer they're in the master `.gitignore` but not in the `task/rewrite` branch `.gitignore` that we're working off (yet)

kale-stew:
My bad 🙂 Thank you for adding those

test.mdx
```diff
@@ -0,0 +1,20 @@
+# Hello, _world_!
+
+Below is an example of JSX embedded in Markdown. <br /> **Kara and dom were successful!**
```
kale-stew: 🎉

package.json
```diff
@@ -22,25 +23,41 @@
     "type": "git",
     "url": "https://github.com/FormidableLabs/spectacle.git"
   },
-  "dependencies": {},
+  "browserslist": "> 0.25%, not dead",
+  "dependencies": {
+    "webpack": "^4.33.0",
```
ryan-roemer: these should be devDeps

dist/index.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
```
kale-stew: If `dist/` is being gitignored, should this still be showing up in the diff?

ryan-roemer:
Thanks for ignore file @VirtualDOMinic !  i think the `dist/**` files still need to get manually removed too...

dist/index.html
VirtualDOMinic: Yep, they were still in the git cache, but I think my latest commit (`df726bd`) has potentially solved it! We'll get there in the end 😬

VirtualDOMinic:
@ryan-roemer are you okay with us merging this in its current state? 😄

PR #691, opened 2019-06-14
Rewrite slide display and transition logic.

kiraarghy:
This branch rewrites the core state management logic and animation logic. 

The design philosophy behind this has been moving away from the top-down redux model to a more bottoms-up local-state model (though using reducers as I think they're easier to make sense of with state changes.)

This also implements a move to using React-Spring, which should allow users to implement their own artisan animations between slides, and between slide elements. 

As it is this represents an MVP before moving onto the display and layouts API for slides.


kiraarghy:
Pushed up most recent commits, had to go home due to migraine :(. Will hopefully look to documenting and fixing some issues with the current set-up but hopefully gives some insight into what's happening!

src/hooks/useSlide.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+    if (e.key === 'ArrowRight') {
+      dispatch('prev slide element');
+    }
+  };
+
+  React.useEffect(() => {
+    window.addEventListener('keydown', e => handleKeyDown(e));
+    return function cleanup() {
+      window.removeEventListener('keydown', handleKeyDown());
```
narinluangrath: Could you change to `handleKeyDown` instead of `handleKeyDown()`?

src/hooks/useDeck.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+/**
+ * Exposes DeckContext for use in Deck component and its children
+ */
+
+export const DeckContext = React.createContext();
+
+function useDeck(initialState) {
```
narinluangrath: We don't have to do this now, but eventually we'll want some logic for handling the end of the slides. Maybe we could add another argument `numSlides`. Then we could do `{ currentSlide: (state.currentSlide + 1) % 3 }` to wrap around to the first slide. Just a thought 🤷‍♀

src/components/Slide.js
```diff
@@ -0,0 +1,57 @@
...
+    : props.children;
+  return (
+    <div
+      style={
+        isActive ? { backgroundColor: 'blue' } : { border: '2px solid red' }
+      }
+    >
+      <p>{slideNo}</p>
+      <SlideContext.Provider
+        value={useSlide(initialState, isActive, numberOfSlideElements)}
```
narinluangrath: Should we put some logic to advance to the next slide if we're done "showing" all the slide elements? I was thinking something like,

```js
const [{  currentSlideElement }] = useSlide(initialState, isActive, numberOfSlideElements); // move this to top of component
const [ _, dispatch ] = useContext(DeckContext);

useEffect(() => {
  if (currentSlideElement >= numberOfSlideElements) {
    dispatch('next slide')
  }
  if (currentSlideElement < 0) {
    dispatch('prev slide')
  }
}, [currentSlideElement])

File: src/components/Controls.js [hunk 4]
```diff
@@ -0,0 +1,20 @@
...
+ * Controls navigation through deck
+ * Provided as its own component to easily allow people to add their own alternative
+ */
+
+const Controls = () => {
+  const [state, dispatch] = React.useContext(DeckContext);
+
+  return (
+    <>
+      <button onClick={() => dispatch({ type: 'prev slide' })}>Prev</button>
```
narinluangrath: Would love to move these type names to a variable. We can do this later though.

narinluangrath:
This looks great! Let me know what you think about my comments 😃

src/hooks/useDeck.js
VirtualDOMinic: We're thinking it could be good to have a `loop` (or `noLoop`, depending on our default) prop for the deck, where people can decide whether they want it to loop back to the start or to just prevent any change

src/components/Slide.js
kiraarghy: This should be covered by src/hooks/useSlide.js now I think :)

src/hooks/useDeck.js
narinluangrath: Sounds good to me

narinluangrath:
I think this guy is ready to be merged :)

src/hooks/useSlide.js
```diff
@@ -1,44 +1,113 @@
...
- *
+ * Performs logic operations for all of the slide domain level.
+ * slideElements are dynamic elements within the slide that may
+ * appear/ move around etc.
+ * If remaining elements in slide, these are brought in one by one.
+ * If not, we tell the deck to take us to the next slide.
  */
 
+//  Keeps previous slide number for comparison later.
+let prevSlideNum;
```
narinluangrath: I think using `usePrevious` might be more idiomatic. https://usehooks.com/usePrevious/
We might also run into problems with this global variable if we have multiple invocations of `useSlide`.

src/components/SlideElementWrapper.js
```diff
@@ -1,38 +1,45 @@
...
-      if (transitionEffect) {
-        console.log(transitionEffect.to);
-        set({
-          to: transitionEffect.to
-        });
-      }
-      set({
-        opacity: 1
-      });
+      set({ from: transitionEffect.from, to: transitionEffect.to });
```
narinluangrath: Could we simplify this even further and drop the `from` case and just do `set({ transitionEffect.to })`? Seems to produce the same result I think.

src/components/Slide.js
```diff
@@ -15,12 +15,18 @@ const Slide = props => {
   const { children, slideNum } = props;
 
   const numberOfSlideElements = Array.isArray(children)
-    ? children.filter(x => isComponentType(x, 'SlideElementWrapper')).length
+    ? children.filter(x => isComponentType(x, 'SlideElementWrapper')).length + 1
```
narinluangrath: What's the `+ 1` for?

src/components/Deck.js
```diff
@@ -65,11 +65,13 @@ const Deck = ({ children, loop, keyboardControls }) => {
...
-    state.currentSlide,
-    p => p,
-    filteredChildren[state.currentSlide].props.transitionEffect || defaultSlideEffect
-  );
+  const transitions = useTransition(state.currentSlide, p => p, {
+    ...(filteredChildren[state.currentSlide].props.transitionEffect ||
+      defaultSlideEffect),
+    config: { precision: 0 },
+    unique: true,
+    immediate: rest.animationsWhenGoingBack ? false : state.immediate
```
narinluangrath: This is a nit picky comment but could we pass `animationsWhenGoingBack` as an argument to `useDeck`? It might be nice to put all the "immediate" logic in one place. Not a big deal though, feel free to ignore.

src/components/Slide.js
kiraarghy: Ummmm I'm gonna remove it, thanks for picking it up!

src/components/Deck.js
kiraarghy: Yeaaah this is a great suggestion!

PR #692, opened 2019-06-14
Bug/issue 661 unmounting bug

### LINKED ISSUE #661: No-op when slideshow is programmatically unmounted
### Describe Your Environment

What version of Spectacle are you using? 5.4.0

What version of React are you using? 16.8.3

What browser are you using? Chrome 72

What machine are you on? MacBook Pro 6,1

### Describe the Problem

I am using Spectacle for a simple autoplaying slideshow that should disappear as soon as a user clicks or taps the browser window, to be replaced with a survey the user can take. The slideshow should reappear after *n* seconds of user inactivity.

To manage this user behavior I am using react-idle-timer and conditional rendering based on state. When the user is idle, a `<SlideshowContainer />` returns a `<Slideshow />`, which is the component containing the Spectacle Deck and Slides. When the user becomes active, the `<SlideshowContainer />` returns `null`. 

In practice, when the user becomes active, the slideshow does disappear and the survey appears, but I get the following output to the console: 

`Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.`

    in Manager (created by Connect(Manager))

    in Connect(Manager) (created by Deck)

    in Context (created by Controller)

    in Controller (created by Deck)

    in Provider (created by Deck)

    in Deck (at Slideshow.js:24)

    in Slideshow (at SlideshowContainer.js:9)

Then, after the user becomes idle again, the slideshow reappears but is obviously broken, with slide transitions happening simultaneously or in the wrong timing... or something. I can't quite identify what the exact behavior is.

I was able to achieve a basic version of my desired functionality by using another slideshow library, but I want to be able to use Spectacle because it has other features I need.

I have created a very basic version of my project that reproduces the issue, which can be accessed at this repo:

https://github.com/glassblowerscat/spectacle-bug-demo

Thanks!

narinluangrath:
### Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Another one liner fix. We use `setInterval` to pan through the slides automatically (if the user sets the `autoplay` and `autoplayOnStart` props). We just forgot to run `clearInterval` if the component unmounts. Consequently, `_nextSlide` keeps getting run even when the Deck is unmounted, leading to some weird behavior if Deck gets mounted again.

Fixes #661

#### Type of Change

### How Has This Been Tested?

You can reproduce the problem here: https://github.com/FormidableLabs/spectacle/issues/661
To prove the fix worked I did the following:
- cd into `spectacle-bug-demo`
- change the package.json to `"spectacle": "FormidableLabs/spectacle#bug/issue-661-unmounting-bug"`
- cd into `node_modules/spectacle`
- `yarn install` && `yarn build`
- startup the demo and the bug is fixed

I also added a test to Manager.js

### DIFF
--- a/src/components/manager.js
+++ b/src/components/manager.js
@@ -213,6 +213,9 @@ export class Manager extends Component {
 
   componentWillUnmount() {
     this._detachEvents();
+    if (this.autoplayInterval) {
+      clearInterval(this.autoplayInterval);
+    }
   }
 
   _attachEvents() {
--- a/src/components/manager.test.js
+++ b/src/components/manager.test.js
@@ -14,6 +14,9 @@ const _mockContext = function(slide, routeParams) {
       controls: {},
       progress: {
         pacman: []
+      },
+      autoplay: {
+        pause: {}
       }
     },
     store: {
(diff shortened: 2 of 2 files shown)


glassblowerscat:
🙌

PR #695, opened 2019-06-20
Feature/add presentation hook

narinluangrath:
I branched off of the current HEAD of `rewrite/add-deck`. I can't really demonstrate the functionality of the hook yet because of some bugs in `rewrite/add-deck`. Once those are resolved, I'll merge pull those changes into this PR and finish off the presentation code.

### DIFF
--- a/src/hooks/usePresentation.js
+++ b/src/hooks/usePresentation.js
@@ -1,0 +1,78 @@
+import { useEffect, useState, useCallback, useMemo, useRef } from 'react';
+
+function usePresentation(handleMessage) {
+  const [connection, setConnection] = useState(null);
+  const requestRef = useRef(null)
+  const [errors, setErrors] = useState([]);
+
+  // Open to suggestions for better error handling
+  const addError = e => setErrors(es => ([...es, e]));
+
+  const terminateConnection = useCallback(() => {
+    if (connection) {
+      connection.terminate();
+    }
+  }, [connection])
+
+  // Create a presentation request and store it as a ref
+  useEffect(() => {
+    if (PresentationRequest) {
+      if (!requestRef.current) {
+        requestRef.current = new PresentationRequest(['/']);
+      }
+    } else {
+      addError(new Error('Browser does not support Presentation API'));
+    }
+    return terminateConnection;
+  }, [connection])
+
+  // Listen for messages and call handleMessage
+  useEffect(() => {
+    const handleConnectionList = list => {
+      list.forEach(connection => {
+        connection.onmessage = ({ data }) => handleMessage(data);
+      })
+    };
+    const receiver = navigator && navigator.presentation && navigator.presentation.receiver;
+    if (reciever) {
+      receiver
+        .connectionList
+        .then(handleConnectionList)
+        .catch(addError)
+    } else {
+      addError(new Error('Browser does not support Presentation API'))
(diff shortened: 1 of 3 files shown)


kiraarghy:
Hey @narinluangrath I had a look at this today, with some help from @kitten we managed to get it working and it works pretty well, great work on getting it working with chromecast! 

However we have a suggestion to refactor some of the core state so that we can just send an object over to the presentation window with the state, so maybe something like:

```javascript
sendMessage({
  currentSlide: 1,
  currentSlideElement: 0
});
```

Then that filters through the cast component's state via context.

From my perspective I think that seems to be more of a deterministic patter and may be easier to implement/ maintain, interested in hearing your perspective on this? 😄

src/hooks/usePresentation.js
```diff
@@ -3,25 +3,35 @@ import { useEffect, useState, useCallback, useMemo, useRef } from 'react';
...
-    if (!messageHandlers[key]) {
-      setMessageHandlers({ ...messageHandlers, [key]: handler });
-    }
-  }
+  // add message handler handles adding and removing handlers
+  const addMessageHandler = useCallback((handler, key) => {
+    const keyHandlers = (messageHandlers.current[key] =
+      messageHandlers.current[key] || []);
+    keyHandlers.push(handler);
+    return () => {
```
narinluangrath: I thought only `useEffect` accepts a cleanup function?

src/hooks/usePresentation.js
```diff
@@ -3,25 +3,35 @@ import { useEffect, useState, useCallback, useMemo, useRef } from 'react';
...
   const [errors, setErrors] = useState([]);
 
-  const addMessageHandler = (handler, key) => {
-    if (!messageHandlers[key]) {
-      setMessageHandlers({ ...messageHandlers, [key]: handler });
-    }
-  }
+  // add message handler handles adding and removing handlers
+  const addMessageHandler = useCallback((handler, key) => {
+    const keyHandlers = (messageHandlers.current[key] =
```
narinluangrath: I didn't know you could do that, neat!

narinluangrath:
@kiraarghy Yes! I like that a **lot** better. It will actually make communication easier for both ends, we don't need a `syncedDispatch`. I'm imagining something like

`Deck.js`
```
const [state, dispatch] = useDeck(...);
const { 
  sendMessage, 
  isConnected, 
  startConnection, 
  terminateConnection, 
  isReceiver, 
  addMessageHandler 
} = usePresentation();

// on page load, start the connection
useEffect(() => {
 startConnection();
 return terminateConnection; // cleanup
}, []);

// (controller) when the Deck state changes, trigger a `sendMessage` 
useEffect(() => {
  if (isConnected && !isReceiver) {
    sendMessage(state);
  }
}, [state, isConnected, isReceiver]);

// (receiver) update the Deck state when you receive the message
useEffect(() => {
  if (isConnected && isReceiver) {
    addMessageHandler(newDeckState => {
      // Use the `dispatch` function to update
      // the state. Ideally, add a type to the
      // reducer to accept a full state replacement
    })
  }
}, [isConnected, isReceiver])
```

If we have some state for `Slide.js` we're need to repeat the above code in that file as well.

narinluangrath:
Feel free to merge

PR #704, opened 2019-08-12
MDX Slide Parsing and CLI

SunburtReynolds:
### Description

This PR introduces a two new pieces of functionality:
* MDX file parsing that delineates Spectacle `<Slide />`s by the presence of `---` in the MDX
*  CLI functionality that can instantly boot up a `webpack-dev-server` instance for an arbitrary MDX file

These two features work hand in hand to simplify how MDX files can be used in Spectacle. Like in Old Spectacle, `---` can be used to denote separate slides. This will work separately from the CLI. However, in combination with the CLI, booting up a hot-reloading Spectacle presentation from an MDX file is very straightforward. Once Spectacle is installed globally (or `npm link`'d from the cloned repo), a user can run this from any arbitrary directory:
`spectacle -s <name of file>`.

The `-s` flag tells the CLI that the user wants to boot up a presentation from the provided file. The CLI defaults to looking for a file named `slides.mdx`, but the user can give an alternative name/path. For example:
`spectacle` (looks for `slides.mdx` in the current directory) or `spectacle -s ./nested/my-pres.mdx` (points Spectacle directly to the MDX file). In the future, other file types might be supported with the `-s` option. For now, only MDX files are supported.

**Webpack** - when you see the webpack configuration, you might wonder why it requires so many changes. It appears that when webpack is run from a directory outside of the Spectacle project (i.e. as a globally installed CLI tool), it has immense trouble finding the loaders and babel tools that are used in the configuration. The only solution that I found to bypass this problem was to use Node's `path` library to construct paths to the dependencies.

**MDX Loader** - the loader file has comments to explain each step. It takes significant inspiration from the custom MDX loader in Old Spectacle. However, it includes updates that are a necessity born of the `mdx-js` version that we're using.

ezgif com-video-to-gif-2

#### Type of Change

- New feature (non-breaking change which adds functionality)
- This change requires a documentation update

### How Has This Been Tested?

To test new functionality:
1. check out this branch locally
2. run `npm link` in the spectacle directory
3. move to any directory you like, e.g. `~/Documents/test/nest`
4. create a new MDX file with valid content (see https://github.com/FormidableLabs/spectacle/pull/704/files#diff-5e02b1a431b9e43c86f7a48c7a90294a as an example); remember that you must create any components that are imported into the MDX
5. run `spectacle -s <your mdx file>` to boot up the Spectacle presentation


bin/args.js
```diff
@@ -0,0 +1,33 @@
...
+
+  return Promise.resolve({ mdx });
+};
+
+const args = () =>
+  yargs
+    .usage(`Usage: spectacle -m <file>`)
+
+    // MDX File
+    .option('mdx', {
```
rgerstenberger: This looks funky to me - `Usage: spectacle -m <file>`. As a spectacle consumer, I would assume that `-m` in that scenario means "file name" but here it's actually `mdx`. A better structure would be to have a filename option like `-s/src` which takes a file of either `.js(x)` or `.md(x)`. And then either infer the filetype from webpack or have a boolean flag `-m/mdx` which sets the filetype to `md(x)`.

It's also weird that the default filename is hidden behind a flag. I would assume you could just run `spectacle` and have smart defaults like "look for slides.js and spin up webserver"

rgerstenberger:
I'm curious as to why this project has mixed case filenames and directories. My understanding was that we try to use `kabob-case` for file paths because some OSes are case insensitive leading to crazy path bugs.

bin/args.js
SunburtReynolds: What would you expect `spectacle -s <jsx file name here>` to do? AFAIK that's undefined behavior.

rgerstenberger: Run a slideshow from JSX slides. It's currently undefined yes. I think the main point is that the slide file is separate from "is this mdx". And that decision can happen from a user supplied flag, OR we might be able to infer it based on the source file type.

SunburtReynolds:
@rgerstenberger here's another PR to address the `kebab-case` comment!
https://github.com/FormidableLabs/spectacle/pull/705

bin/args.js
SunburtReynolds: Makes sense to me. Here's the decision tree I've got in mind:

Did the user supply a file?

If no, default to (?) `slides.mdx`.

If yes, what kind of file did they supply an mdx file or a jsx file?

If mdx, boot up the presentation.

If jsx (or other), console log about currently unsupported behavior.

rgerstenberger: Sounds reasonable, though I'd eventually expect the default to be `slides.jsx`

rgerstenberger:
Looks good to me!

webpack.config.js
```diff
@@ -1,6 +1,42 @@
 const path = require('path');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 
+/*
```
narinluangrath: @SunburtReynolds I think the `context` config could fix this: https://webpack.js.org/configuration/entry-context/#context

SunburtReynolds: @narinluangrath unfortunately, I tried `context` but could not get it to work! If you have the time to try it out and find a particular configuration that doesn't require all the absolute paths to dependencies, I will absolutely take it 🙏

narinluangrath: Aw man, that is really surprising. I'll try it out tonight.

PR #705, opened 2019-08-12
use kebab-case consistently throughout project

SunburtReynolds:
### Description

As per @rgerstenberger's comment on another PR, we want to use `kebab-case` consistently to avoid cross-platform file system issues.

#### Type of Change

- Bug fix (non-breaking change which fixes an issue)

### How Has This Been Tested?

I retested CLI functionality, and everything is normal.


narinluangrath:
LGTM

examples/MDX/test-mdx.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import slides from './slides.mdx';
-import Deck from '../../src/components/Deck';
-import Slide from '../../src/components/Slide';
+import Deck from '../../src/components/deck';
```
rgerstenberger: This is still `/MDX/*` which doesn't solve the problem.

SunburtReynolds: It's lower case locally. I tried to make it upper then lower case again. No diff in git.

Screen Shot 2019-08-12 at 3 22 34 PM

SunburtReynolds: Fixed!

PR #706, opened 2019-08-14
Update dependencies

### LINKED ISSUE #720: Support React Hooks
Spectacle seems to be using React 16.5.2. Hooks were introduced in React 16.8. Is it possible to use a higher React version than the default one that come with the spectacle starter project?
Hooks will simply creating stateful and interactive presentations.

kale-stew:
### Description 

This updates a handful of dependencies that are long out of date. It does not include a few packages that require a bit more effort to update:

- emotion
- react-live
- react-redux
- eslint

Closes #720 

### How Has This Been Tested?

With our tests, but also with some manual feature testing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "redux": "^4.0.1",
     "redux-actions": "^2.6.4",
     "to-style": "^1.3.3",
-    "victory-core": "^31.1.0"
+    "victory-core": "^33.0.6"
   },
   "peerDependencies": {
     "prop-types": "^15.6.2",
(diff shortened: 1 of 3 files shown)


philipwhiuk:
Was this abandoned because of time constraints or is there some vicious problem here? I get GitHub dependency spammed due to a Spectacle deck - would be nice to pull it back to the bleeding edge.

kale-stew:
@philipwhiuk what dependency spam are you seeing when using Spectacle? I am rolling back the changes in this PR to isolate updates to non-problematic upgrades (even eslint seems unhappy with us now), but I want to make sure I'm including the dependencies that are causing your issue.

philipwhiuk:
Sorry I was probably mistaken. I took another look at my `yarn audit` output. Spectacle itself seems okay. The renderer however I've raised https://github.com/FormidableLabs/spectacle-renderer/pull/10 to fix :)

PR #708, opened 2019-08-21
Fix Transitions and State Updates

carlos-kelly:
*Problem Statement*
Our animations are run-to-completion spring-based, therefore frequent firing of navigation commands will cause React to unmount a currently animating element. To maintain the existing animation library and public API we need to prevent the app from firing off a navigation event while an animation is underway.

*Solution*
Use Wonka to run transitions and batch them. This PR also fixes the backwards transition of elements when the deck is navigating in the reverse direction.


src/components/Deck.js
```diff
@@ -18,65 +19,99 @@ import { useTransition, animated } from 'react-spring';
...
-    i // eslint-disable-next-line react/display-name
-  ) => ({ style }) => (
-    <animated.div style={{ ...style }}>
-      {{
-        ...x,
-        props: { ...x.props, slideNum: i, keyboardControls }
-      }}
-    </animated.div>
-  ));
+  const slideElementMap = React.useMemo(() => {
```
carlos-kelly: We use an element map because for going backwards we need to know how many elements there are to remove them in a backwards-fashion. This also lets us fix bounds issues too.

src/components/Deck.js
```diff
@@ -18,65 +19,99 @@ import { useTransition, animated } from 'react-spring';
...
+    slideElementMap
   );
+  const userTransitionEffect =
+    filteredChildren[state.currentSlide].props.transitionEffect || {};
+  const transitionRef = React.useRef();
+
+  React.useEffect(() => {
+    const runSlideTransition = () =>
+      transitionRef.current.start().then(() => signal());
+    if (firstRun) {
```
carlos-kelly: We use a firstRun flag because the first dispatch is based on page load, not key press so we need that first animation call to be synchronized with the mutex.

src/components/SlideElementWrapper.js
```diff
@@ -1,41 +1,74 @@
...
+    { reverseDirection, currentSlideElement, immediateElement }
+  ] = React.useContext(DeckContext);
+  const { signal } = React.useContext(AnimationMutexContext);
+  const activeElement = elementNum === currentSlideElement;
+  const upcomingElement =
+    (elementNum > currentSlideElement && !reverseDirection) ||
+    (elementNum < currentSlideElement && reverseDirection);
+  const previousElement =
+    (elementNum < currentSlideElement && !reverseDirection) ||
+    (elementNum > currentSlideElement && reverseDirection);
```
carlos-kelly: This component got a little more complex because we need to handle all the edge cases for slides when going forwards or backwards—or when direction changes midway.

React-spring does not commit changes in a permanent style to their nodes, so we have to know if they have already animated in or not (previousElement) so we can set the style props manually. Otherwise we would lose transitions on next element navigation. The conditions for upcoming and previous elements are reversed depending on the direction of the navigation.

src/components/SlideElementWrapper.js
```diff
@@ -1,41 +1,74 @@
...
 
-  const [styleProps, set] = useSpring(() => transitionEffect.from);
+  const [styleProps, set] = useSpring(() => {
+    if ((activeElement || upcomingElement) && !previousElement) {
+      return reverseDirection ? transitionEffect.to : transitionEffect.from;
+    } else if (previousElement) {
```
carlos-kelly: If the element is active (aka about to animate in) or is a future element and we are going forwards we want the transition effect to be "from", because that's the starting state. If we're going in reverse direction, then we want it to be to, since elements start as shown when going reverse.

src/components/SlideElementWrapper.js
```diff
@@ -1,41 +1,74 @@
...
+    (elementNum > currentSlideElement && !reverseDirection) ||
+    (elementNum < currentSlideElement && reverseDirection);
+  const previousElement =
+    (elementNum < currentSlideElement && !reverseDirection) ||
+    (elementNum > currentSlideElement && reverseDirection);
 
-  const [styleProps, set] = useSpring(() => transitionEffect.from);
+  const [styleProps, set] = useSpring(() => {
+    if ((activeElement || upcomingElement) && !previousElement) {
+      return reverseDirection ? transitionEffect.to : transitionEffect.from;
```
carlos-kelly: If the element is active (transitioning in) or upcoming then we want the spring starting point to be the from state which is the not visible state. This is opposite in reverse direction since elements start visible then animate out.

src/components/SlideElementWrapper.js
```diff
@@ -1,41 +1,74 @@
...
+    } else if (previousElement) {
+      return reverseDirection ? transitionEffect.from : transitionEffect.to;
```
carlos-kelly: For previousElements, that means elements that have already animated in for forwards direction, or elements that have disappeared in reverse direction, we want to persist their completed state transitions since react-spring doesn't do that automatically.

src/components/SlideElementWrapper.js
```diff
@@ -1,41 +1,74 @@
...
+    } else {
+      return transitionEffect.to;
```
carlos-kelly: This else case is a fallback since react spring throws an exception if no transition configuration is hit. But in regular operation this case should not hit, but if it does, just make the element show rather than blow up the app.

src/components/SlideElementWrapper.js
```diff
@@ -1,41 +1,74 @@
...
-  // if so trigger transition, if not then to initial!
   React.useEffect(() => {
-    if (state && elementNum <= state.currentSlideElement) {
+    if (activeElement) {
       set({
-        ...transitionEffect,
-        immediate: state.immediate
+        ...transitionEffect.to,
+        immediate: immediateElement,
+        onRest: () => signal()
```
carlos-kelly: When an element transition is complete we signal the mutex to release the lock. If there are any queued transitions, then they will execute.

src/hooks/useDeck.js
```diff
@@ -18,35 +24,96 @@ function useDeck(initialState, numSlides, looping, animationsWhenGoingBack) {
...
-          immediate: animationsWhenGoingBack ? true : false
+          currentSlideElement: Math.max(
+            slideElementMap[state.currentSlide - 1],
+            0
+          ),
+          immediate: !!animationsWhenGoingBack,
+          immediateElement: true,
+          reverseDirection: true
+        };
+      case 'NEXT_SLIDE_ELEMENT': {
```
carlos-kelly: We add the element navigation to the deck reducer because it reduces the number of dispatches and keeps all navigation in a unified location. This will be key when we integrate URL-based routing. Aka `spectacle:3000/1/2` meaning slide 1, element 2. Tracking navigation of slides and their elements in separate reducers makes them individually smaller, but more challenging to keep data in sync.

src/hooks/useDeck.js
```diff
@@ -18,35 +24,96 @@ function useDeck(initialState, numSlides, looping, animationsWhenGoingBack) {
...
         };
+      }
       default:
         return { ...state };
     }
   }
-  const [state, dispatch] = React.useReducer(reducer, initialState);
+  const [state, dispatch] = React.useReducer((state, action) => {
+    const _reducer = reducer(state, action);
+    // console.log(_reducer);
```
carlos-kelly: This is logger middleware for the time being and is useful for development and debugging. It will be removed come release time.

src/hooks/useMutex.js
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+
+function useMutex() {
```
carlos-kelly: A mutex is used to keep navigation from unmounting element and slide undergoing animations causing a state update exception. React spring does not have the concept of time (as it is a spring-based animation library) so the animations are run-to-completion.

src/hooks/useMutex.js
```diff
@@ -0,0 +1,50 @@
...
+  const queue = React.useRef([]);
+  const lock = React.useRef(false);
+
+  const wait = React.useCallback(() => {
```
carlos-kelly: `wait` is an operation that looks at the execution queue and sees if there are any tasks that need to be run. If there are, the lock is acquired and the task is executed.

src/hooks/useMutex.js
```diff
@@ -0,0 +1,50 @@
...
+    const next = queue.current.shift();
+    if (!next) {
+      return;
+    }
+    lock.current = true;
+    return Promise.resolve(next());
+  }, []);
+
+  const signal = React.useCallback(() => {
```
carlos-kelly: `signal` is an operation that lets the mutex know the operation is complete and to release the task. On signal complete, we call `wait` to see if there is another task in the queue that was added while the previous task was executing.

src/hooks/useMutex.js
```diff
@@ -0,0 +1,50 @@
...
+    lock.current = false;
+    return wait();
+  }, [wait]);
+
+  const synchronize = React.useCallback(
+    runnable => {
```
carlos-kelly: `synchronize` is the operation that adds a task to the queue. It will execute the task if the lock is not acquired.

src/hooks/useMutex.js
```diff
@@ -0,0 +1,50 @@
...
+    [wait, lock]
+  );
+
+  return { synchronize, signal, wait };
+}
+
+const AnimationMutexContext = React.createContext();
+
+function AnimationProvider(props) {
+  const mutex = useMutex();
```
carlos-kelly: This provider's value is a way for different components to access the same animation mutex.

src/hooks/use-transition-pipe.js
```diff
@@ -0,0 +1,23 @@
...
+import PropTypes from 'prop-types';
+import { pipe, subscribe, makeSubject } from 'wonka';
+
+const TransitionPipeContext = React.createContext({});
+
+function TransitionPipeProvider(props) {
+  const [transitionSource, runTransition] = React.useRef(makeSubject()).current;
+  pipe(
+    transitionSource,
+    subscribe(transition => transition.start())
```
kitten: Small nit 😅  we're never unsubscribing here, so if this provider ever updates you'll see duplicate transitions potentially.

Maybe it's worth just wrapping the entire subject and subscription in a useMemo with no deps? We won't have to unsubscribe since no consumers can access runSubscription if this unmounts

carlos-kelly:
Oh great point. I like your idea. Otherwise I’d call unsubscribe in the clean up or a useEffect.

PR #717, opened 2019-09-04
Base Theming

carlos-kelly:
This PR adds base theming via Styled System and Styled Components plus a handful of Text components like Header and List.

Each component is a styled component with themed defaults. Meaning the user can either customize it with styled system props `<Text fontSize />` or extend it using Styled Components directly.

Styled theme props can either take the name of a theme label or a raw value
`<Text fontSize="big" />` or `<Text fontSize="24px" />`. This PR has a default theme based on Formidable colors.

Screen Shot 2019-09-04 at 12 34 40 PM


src/components/deck.js
```diff
@@ -111,7 +114,15 @@ const Deck = ({ children, loop, keyboardControls, ...rest }) => {
...
     document.body.style.margin = '0';
-  }, []);
+    document.body.style.background =
+      themeContext.colors[rest.backgroundColor] ||
+      rest.backgroundColor ||
+      themeContext.colors.tertiary;
+    document.body.style.color =
+      themeContext.colors[rest.textColor] ||
+      rest.textColor ||
+      themeContext.colors.primary;
```
carlos-kelly: We want to set the body color to the theme background and text color to theme text color. We can source the theme from either
1. The theme label
2. A raw value
3. The default theme value

src/components/index.js
```diff
@@ -0,0 +1,25 @@
...
+  Slide,
+  SlideElementWrapper,
+  CodePane,
+  OrderedList,
+  Quote,
+  Heading,
+  ListItem,
+  UnorderedList,
+  Text
+};
```
carlos-kelly: This unifies all of our exports into a single point.

src/theme/default-theme.js
```diff
@@ -0,0 +1,21 @@
...
+    header: '48px',
+    subHeader: '32px',
+    text: '18px'
+  },
+  space: {
+    headerMargin: '16px',
+    textMargin: '16px',
+    listMargin: '16px 0'
+  }
+};
```
carlos-kelly: These are some defaults for our theme's space, colors, and fonts. https://styled-system.com/getting-started

rgerstenberger:
Does theming allow for adding a logo (say Formidable's ;)) somewhere on all slides (say bottom right)?

src/theme/default-theme.js
```diff
@@ -0,0 +1,21 @@
+export default {
+  colors: {
+    primary: '#ebe5da',
+    secondary: '#f04d21',
+    tertiary: '#1e2852'
+  },
+  fonts: {
+    header: 'Helvetica Neue',
```
rgerstenberger: Looks like this should include fallback fonts?

src/components/typography.js
```diff
@@ -0,0 +1,82 @@
...
+Text.defaultProps = {
+  color: 'primary',
+  fontFamily: 'text',
+  fontSize: 'text',
+  textAlign: 'left',
+  padding: 0,
+  margin: 'textMargin'
+};
+
+const Heading = styled(Text)({});
```
SunburtReynolds: Do you have to include `({})`?

src/theme/default-theme.js
SunburtReynolds: Do you think we should make our theme public/available to the public?

src/components/typography.js
carlos-kelly: I have to do something either `({})` or back-ticks. IMO, a blank object looks less strange.

src/theme/default-theme.js
carlos-kelly: Yes! I plan on creating a helper that lets you use or extend it, which really is Object.assign. But still a helper.

PR #719, opened 2019-09-09
Templates, Fix to a 16:9 aspect ratio.

carlos-kelly:
### Description

This fixes a slide to a 16:9 ratio which fits most TVs and helps unify the presenter display view.

Screen Shot 2019-09-09 at 11 29 16 AM

Template Support:

Screen Shot 2019-09-09 at 2 56 44 PM

### DIFF
--- a/examples/js/test-js.js
+++ b/examples/js/test-js.js
@@ -94,7 +94,7 @@ const TestJs = () => (
       >
         {Array(9)
           .fill('')
-          .map(index => (
+          .map((_, index) => (
             <FlexBox key={`formidable-logo-${index}`}>
               <Image src={formidableLogo} width={100} />
             </FlexBox>
--- a/examples/js/test-js.js
+++ b/examples/js/test-js.js
@@ -39,7 +39,7 @@ const TestJs = () => (
         width={1}
       >
         <Text fontSize={12}>
-          Slide {slideNumber + 1} of {numberOfSlides}
+          Slide {slideNumber} of {numberOfSlides}
         </Text>
         <Box padding={10}>
           <Image src={formidableLogo} width={100} />
--- a/examples/js/test-js.js
+++ b/examples/js/test-js.js
@@ -22,16 +22,32 @@ const cppCodeBlock = `#include <iostream>
 int main()
 {
   auto curried_add = [](int x) -> function<int(int)> { return [=](int y) { return x + y; }; };
-  
+
   auto answer = curried_add(7)(8);
   std::cout << answer << std::endl;
-  
+
   return 0;
 }`;
 
 const TestJs = () => (
-  <Deck>
-    <Slide slideNum={1}>
+  <Deck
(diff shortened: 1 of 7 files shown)


examples/js/test-js.js
```diff
@@ -22,16 +22,32 @@ const cppCodeBlock = `#include <iostream>
...
+      >
+        <Text fontSize={12}>
+          Slide {slideNumber + 1} of {numberOfSlides}
+        </Text>
+        <Box padding={10}>
+          <Image src={formidableLogo} width={100} />
+        </Box>
+      </FlexBox>
+    )}
+  >
```
carlos-kelly: Components we use inside the template automatically consume the used theme.

examples/js/test-js.js
```diff
@@ -22,16 +22,32 @@ const cppCodeBlock = `#include <iostream>
...
   auto answer = curried_add(7)(8);
   std::cout << answer << std::endl;
-  
+
   return 0;
 }`;
 
 const TestJs = () => (
-  <Deck>
-    <Slide slideNum={1}>
```
carlos-kelly: We no longer need these as we add them via props based on their order in the tree.

examples/js/test-js.js
```diff
@@ -22,54 +22,74 @@ const cppCodeBlock = `#include <iostream>
...
+  <Deck
+    template={({ numberOfSlides, slideNumber }) => (
+      <FlexBox
+        justifyContent="space-between"
+        position="absolute"
+        bottom={0}
+        width={1}
+      >
+        <Text fontSize={12}>
+          Slide {slideNumber + 1} of {numberOfSlides}
```
SunburtReynolds: If we want the slide # to match the url, then it should be 0 indexed. Might be confusing if they change the url manually

examples/js/test-js.js
```diff
@@ -22,54 +22,74 @@ const cppCodeBlock = `#include <iostream>
...
+    )}
+  >
+    <Slide>
       <Heading>Spectacle</Heading>
       <Text>Hello There 🤗</Text>
       <Quote>This is a Formidaquote!</Quote>
       <OrderedList>
-        <ListItem>This is an</ListItem>
-        <ListItem>ordered list</ListItem>
+        <Appear elementNum={1}>
```
SunburtReynolds: FYI, elem num should be 0 indexed as well.

examples/js/test-js.js
carlos-kelly: Good point!

src/components/slide.js
```diff
@@ -5,14 +5,27 @@ import { DeckContext } from '../hooks/use-deck';
 import styled from 'styled-components';
 import { color } from 'styled-system';
 
+const SlideContainer = styled('div')`
+  ${color};
+  position: relative;
+  width: 100%;
+  padding-top: 56.25%;
```
SunburtReynolds: Nice.

PR #724, opened 2019-09-13
Presenter mode slide preview

SunburtReynolds:
This PR adds the active and next slide previews to the presenter deck. It will stay in sync with the slide elements shown in the associated audience deck.

Thanks to @carlos-kelly 's awesome work, the aspect ratio dictated by the user (or the default) will be maintained within the slide previews.

Screen Shot 2019-09-12 at 6 58 40 PM


0xdabbad00:
How do these stay in sync? I have one browser window opened to http://localhost:3000/?presenterMode=true and the other to http://localhost:3000/ and they do not stay in sync when I transition between slides.

0xdabbad00:
I figured it out. I had to use an external monitor.

PR #726, opened 2019-09-17
Add Progress and Full Screen components.

carlos-kelly:
Add Progress and Full Screen components. Update the default template to use those.

*Windowed Mode*
Screen Shot 2019-09-17 at 3 50 36 PM

*Full Screen Mode*
Screen Shot 2019-09-17 at 3 50 43 PM

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,8 +2,8 @@ import React from 'react';
 import { render } from 'react-dom';
 
 // START: test components to try rendering:
-// import TestMDX from './examples/mdx/test-mdx';
-import TestJs from './examples/js/test-js';
+import TestMDX from './examples/mdx/test-mdx';
+// import TestJs from './examples/js/test-js';
 // END: test components to try rendering
 
 /**
@@ -13,4 +13,4 @@ import TestJs from './examples/js/test-js';
  * to hot-reload with new contents.
  */
 
-render(<TestJs />, document.getElementById('root'));
+render(<TestMDX />, document.getElementById('root'));
--- a/src/components/deck/index.js
+++ b/src/components/deck/index.js
@@ -176,7 +176,6 @@ const Deck = ({
       const staticSlides = filteredChildren.map((slide, index) =>
         React.cloneElement(slide, {
           slideNum: index,
-          numberOfSlides,
           template: rest.template
         })
       );
--- a/src/components/deck/index.js
+++ b/src/components/deck/index.js
@@ -92,7 +92,7 @@ const Deck = ({
   }, [filteredChildren]);
 
   // Initialise useDeck hook and get state and dispatch off of it
-  const { state, dispatch } = useDeck(initialState);
+  const { state, dispatch } = useDeck({ ...initialState, numberOfSlides });
   const themeContext = React.useContext(ThemeContext);
 
   React.useLayoutEffect(() => {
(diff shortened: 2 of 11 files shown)


index.js
```diff
@@ -2,8 +2,8 @@ import React from 'react';
 import { render } from 'react-dom';
 
 // START: test components to try rendering:
-// import TestMDX from './examples/mdx/test-mdx';
-import TestJs from './examples/js/test-js';
+import TestMDX from './examples/mdx/test-mdx';
+// import TestJs from './examples/js/test-js';
```
SunburtReynolds: At some point, we should add an index page that lets you pick which test you want to render.

src/components/deck/index.js
```diff
@@ -92,7 +92,7 @@ const Deck = ({
   }, [filteredChildren]);
 
   // Initialise useDeck hook and get state and dispatch off of it
-  const { state, dispatch } = useDeck(initialState);
+  const { state, dispatch } = useDeck({ ...initialState, numberOfSlides });
```
SunburtReynolds: Why is `numberOfSlides` getting passed to `useDeck`?

src/hooks/use-url-routing.js
```diff
@@ -80,6 +80,19 @@ export default function useUrlRouting(options) {
     [countSlideElements, isSlideElementOutOfBounds, isSlideOutOfBounds]
   );
 
+  const goToSlide = React.useCallback(
```
SunburtReynolds: I'm a little hesitant in exposing this method. The `onHistoryChange` function will correct any boundary errors. But it does seem like a lot of power to lend to the rest of the app.

src/hooks/use-url-routing.js
```diff
@@ -80,6 +80,19 @@ export default function useUrlRouting(options) {
...
+    slideNumber => {
+      const qs = queryString.stringify({
+        presenterMode: currentPresenterMode || undefined,
+        immediate: true,
+        slide: slideNumber,
+        slideElement: 0
```
SunburtReynolds: This should be `DEFAULT_SLIDE_ELEMENT_INDEX`. A value of `0` would show the first `Appear` item instead of giving the presenter a chance to animate it in.

src/components/progress.js
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+import propTypes from 'prop-types';
+import styled from 'styled-components';
+import { DeckContext } from '../hooks/use-deck';
+
+const Circle = styled('figure')`
```
SunburtReynolds: I didn't see `figure` in the styled components API. What does it do?

src/components/fullscreen.js
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
+import propTypes from 'prop-types';
+
+const FullScreen = props => {
+  const toggleFullScreen = React.useCallback(() => {
+    if (!document.fullscreenElement) {
+      document.documentElement.requestFullscreen();
+    } else {
+      if ('exitFullscreen' in document) {
```
SunburtReynolds: I've never seen `in` used like this! Dope.

src/components/deck/index.js
SunburtReynolds: Nvm, I see why. If we want to keep`numberOfSlides` in deck context, can you unify that usage across the app? Some places are getting it via props, some via context.

For instance, https://github.com/FormidableLabs/spectacle/blob/db738e7ec5ab86d8be5f585f5c32edc12a5079c7/src/components/deck/index.js#L178.

Instead, we can proxy `numberOfSlides` through `useSlide` instead of copying it into props as in that link.

PR #730, opened 2019-09-27
[WIP] ESM build

alex-saunders:
**DOING**
- Rollup ESM build
- Rollup CJS build (may as well)

**TODO**
- Alias react to es-react throughout all dependencies (this can be copied from @ryan-roemer importmap-shim work but would be cool to investigate alternatives (for which there might be none?))
- Example `one-page.html` file
- (Proper PR description)


ryan-roemer:
@alex-saunders This all looks good so far! But, I should mention that I'm not experienced much at all with Rollup.

For taking in deps, can we just bundle in all the CJS stuff like `react`, `query-string` in the bundle so we don't need to "do" anything there? Basically, it's a huge fat ESM bundle, but will work straight up (although I do realize this means the `one-page.html` probably needs to match `react` versions and I'm not sure if "two reacts" in an app is a problem)...

kitten:
@ryan-roemer I think I've mentioned this on Slack already, but I'll post a longer reply here 😄 

> For taking in deps, can we just bundle in all the CJS stuff like react, query-string in the bundle so we don't need to "do" anything there

I see that this makes sense for `one-page.html`, but we'd be making things "worse" for two ecosystems here: The usual build/npm ecosystem _and_ the ESM browser import ecosystem.

If we add a build that is distributable on npm, then we'd want the same one to be consistently used, no matter whether it's for Webpack / CRA / our CLI / etc, or whether it's the ESM bundle for the browser.

If we bundle everything into the bundle then we run into problems of stale dependencies, so bundling `react` / `react-dom` / etc is not a good idea, since they're peer dependencies, so that other dependencies can still use them.

It'd really make it harder to distribute the same output for both npm and for ESM browser imports. The current state of this only needs our `es-react` remapping. (And the `es-react` merge + publish, `prism-react-renderer` v1.0.0 is now ready)

So I think all things considered having the build as it is now is perfect :tada: and mergeable already, because it's suitable for npm and usual build pipelines but also browser ESM (apart from React)

(Note: This is also embedding `query-string` already, as that's a rarely changing & safe to bundle dependency that's rather small and that we may not want to find a replacement for _yet_ https://github.com/FormidableLabs/spectacle/pull/730/files#diff-ff6e5f22a9c7e66987b19c0199636480R18)

ryan-roemer:
@kitten -- Don't our transitive dependencies have `react`? Or all they all peerDeps? Have we validated in the bundle that we don't end up with any other "real react"s?

As to what to place in the `dist` bundle, I don't care what's technically in or out, as long as it works out of the box for `one-page.html` with no importmaps, shims or anything else. As long as that's the case, then I'm good -- basically, a working `one-page.html` file in this PR that I open up in a browser that "just works" is my validation criteria ;)

kitten:
@ryan-roemer Yes, our transitive dependencies partly have React (or including any dependencies that a user may include), so our `external` config excludes everything from our bundle (all `dependencies` and `peerDependencies`) except for `query-string` (for now)

We can't get to a point where we can just make it work with `one-page.html` unless we bundle React & React DOM or create a separate bundle for `one-page.html`. But as stated I very much believe that muddies the waters for both browser ESM imports and npm published dependencies. We'd be publishing a bundle that goes against the common practices for both these communitities 😢 

So I'd much rather prefer using the same ESM bundle that's also used by user build pipelines (the `package.json:module` one) for `one-page.html` and rely on an import remapper that replaces `react` with `es-react` until the React team acknowledges that they need to fix ESM.

ryan-roemer:
@alex-saunders -- @kitten has what sounds like a great additional goal of "a general purpose ESM lib distribution". after this PR straight up supports one-page.html, feel free to add that too! (Maybe `dist/one-page.js` for main work and `dist/lib.js` for Phil's idea)

alex-saunders:
@ryan-roemer @kitten Just so I'm clear on the goals here, do we want to produce 2 bundles, where `one-page.js` would work 'out of the box' with no importmaps/shims/etc. and has `react` bundled? Whereas `lib.js` would be a more 'traditional' ESM bundle that, when used, would require `react` to be importmapped to `es-react`?

ryan-roemer:
The main goal for this PR is supporting one-page with a bundle that doesn’t require any import mapping or hacks. Can be UMD or ESM or anything, it just has to work.

Phil can help outline the goals for “another” ESM bundle that you can add here or ticket for later. I wouldn’t worry about it from one-page perspective because as once we have to do an import map/ shim / other hacks we may as well just import map the whole thing like I already have in my other branch without needing any build at all :)

kitten:
@ryan-Roemer just talked with Alex about this and we’re thinking of providing four bundles in this PR:

- CJS bundle without externals (`package.json:main`)
- ESM bundle without externals  (`package.json:module`)
- one-page UMD with all externals embedded
- one-page ESM with all externals embedded

package.json
```diff
@@ -2,18 +2,20 @@
...
+  "module": "dist/spectacle.es.js",
   "types": "index.d.ts",
   "scripts": {
-    "build": "",
     "lint": "eslint --ignore-path .gitignore .",
     "lint-fix": "",
     "prettier": "prettier \"**/*.{js,json,ts,css,md}\"",
     "prettier-fix": "npm run prettier -- --write",
     "deploy": "",
     "start": "builder run --env \"{\\\"BABEL_ENV\\\":\\\"esm\\\"}\" build-wds-base -- --port=3000 --hot",
```
ryan-roemer: TODO: We'll need to rewrite `start` to use rollup and remove `build-wds-base`

package.json
```diff
@@ -58,6 +66,14 @@
...
+    "rollup-plugin-babel": "^4.3.3",
+    "rollup-plugin-buble": "^0.19.8",
+    "rollup-plugin-commonjs": "^10.1.0",
+    "rollup-plugin-node-resolve": "^5.2.0",
+    "rollup-plugin-replace": "^2.2.0",
+    "rollup-plugin-terser": "^5.1.2",
+    "rollup-plugin-unpkg": "^0.0.3",
     "tslint": "^5.12.1",
     "typescript": "^3.2.2",
     "webpack": "^4.40.2",
```
ryan-roemer: ... and remove all webpack deps :)

rollup.config.js
```diff
@@ -16,7 +15,9 @@ if (pkgInfo.peerDependencies)
 if (pkgInfo.dependencies) external.push(...Object.keys(pkgInfo.dependencies));
 
 // TODO: query-string does not have an ESM build, so we'll just bundle and ship it
-external = external.filter(x => x !== 'query-string');
+external = external.filter(
+  x => x !== 'query-string' && x !== 'styled-components'
```
alex-saunders: I'm bundling styled-components as the `package.json:module` field points towards an ESM version compatible with bundlers but **not** browser-ready - `packge.json:browser` points towards what should be a browser-ready ES module version of styled-components but even this includes some non-ES compat dependencies (like `stylis-rule-sheet`) and given that unpkg does not understand this `package.json` field, at the moment it is easier to just bundle styled-components.
Note that this could lead to a stale dependency and we should try and find a workaround (but this is mostly a responsibility of the `styled-components` team).

config/rollup.config.js
```diff
@@ -0,0 +1,212 @@
...
+        esModule: false,
+        file: `./dist/${name}.js`,
+        format: 'cjs'
+      }
+    ]
+  },
+  {
+    ...config,
+    external: id => externalTest(id),
+    plugins: [...makePlugins(), !isProduction && unpkg()].filter(Boolean),
```
alex-saunders: Is there a big downside to including the unpkg URLs in the production build? I've omitted them here but i'm not totally convinced either way.

ryan-roemer:
Hi! I was going to kick the tires on this, but get this error:

```
$ yarn build
yarn run v1.16.0
$ rollup -c config/rollup.config.js

src/components/index.js → ./dist/spectacle.js...
created ./dist/spectacle.js in 3.2s

src/components/index.js → ./dist/spectacle.es.js...
created ./dist/spectacle.es.js in 1.7s

src/components/index.js → ./dist/one-page.umd.js...
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
prism-react-renderer/themes/vsDark (imported by src/components/code-pane.js)
(!) Missing global variable name
Use output.globals to specify browser global variable names corresponding to external modules
prism-react-renderer/themes/vsDark (guessing 'theme$1')
created ./dist/one-page.umd.js in 3.6s

src/components/index.js → ./dist/one-page.js...
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
prism-react-renderer/themes/vsDark (imported by src/components/code-pane.js)
[!] Error: 'Component' is not exported by ../prism-react-renderer/node_modules/react/index.js
https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module
../prism-react-renderer/es/components/Highlight.js (7:16)
5: function _inheritsLoose(subClass, superClass) { subClass.prototype = Object.create(superClass.prototype); subClass.prototype.constructor = subClass; subClass.__proto__ = superClass; }
6: 
7: import React, { Component } from "react";
                   ^
8: import normalizeTokens from "../utils/normalizeTokens";
9: import themeToDict from "../utils/themeToDict";
Error: 'Component' is not exported by ../prism-react-renderer/node_modules/react/index.js
    at error (/Users/rye/scm/fmd/spectacle/node_modules/rollup/dist/rollup.js:10149:30)
    at Module.error (/Users/rye/scm/fmd/spectacle/node_modules/rollup/dist/rollup.js:14132:9)
    at handleMissingExport (/Users/rye/scm/fmd/spectacle/node_modules/rollup/dist/rollup.js:14051:21)
    at Module.traceVariable (/Users/rye/scm/fmd/spectacle/node_modules/rollup/dist/rollup.js:14443:17)
    at ModuleScope.findVariable (/Users/rye/scm/fmd/spectacle/node_modules/rollup/dist/rollup.js:13141:39)
    at Identifier$1.bind (/Users/rye/scm/fmd/spectacle/node_modules/rollup/dist/rollup.js:9219:40)
    at CallExpression$1.bind (/Users/rye/scm/fmd/spectacle/node_modules/rollup/dist/rollup.js:7965:31)
    at CallExpression$1.bind (/Users/rye/scm/fmd/spectacle/node_modules/rollup/dist/rollup.js:11091:15)
    at VariableDeclarator.bind (/Users/rye/scm/fmd/spectacle/node_modules/rollup/dist/rollup.js:7969:23)
    at VariableDeclaration$1.bind (/Users/rye/scm/fmd/spectacle/node_modules/rollup/dist/rollup.js:7965:31)
```

ryan-roemer:
As a separate note, I'd like the ESM version to not use any `es-module-shim`s or other non-standard tags right now. I can already shim everything in my branch without any build, and the shims are not the direction I'm looking for for this particular work.

Thanks!

ryan-roemer:
**Update**: Build works great! (I had a rogue pre-existing yarn link).

I validated the UMD one-page and it looks good! (Here's what I did in case it's helpful: https://github.com/FormidableLabs/spectacle/compare/task/esm-build...task/esm-build-one-page-tweaks)

config/rollup.config.js
kitten: well, so I wouldn't add `isProduction` via `NODE_ENV` in the first place. This is a build setup that should be repeatable in any circumstance, so when we're publishing all these bundles we don't want to deal with running `env NODE_ENV=production yarn build && yarn build` and then maintain different names here all depending on an env flag.

So I'd just remove it and make sure that this config always builds the same set of files, oftentimes production (minified) and unminified bundles.

I think in the case of Spectacle it'd also make sense to just remove the production modes / bundles. There's no real use of `process.env.NODE_ENV` in the code yet and there's no need to actually make Spectacle as small as possible

config/plugins/unpkg-behaviour.js
```diff
@@ -0,0 +1,87 @@
...
+
+    transform(code) {
+      const ast = this.parse(code);
+      const magicString = new MagicString(code);
+
+      walk(ast, {
+        enter(node, parent) {
+          if (node.type === 'Literal' && parent.type === 'ImportDeclaration') {
+            const importLiteral = node.value;
+            const [basePackage, subPackagePath] = importLiteral.split(/\/(.+)/);
```
kitten: This will fail for org packages, i.e. `@test/test`

alex-saunders:
@ryan-roemer @kitten I've got `one-page.js` which is an ESM version with all externals embedded so that no shims or non-standard tags should be required to get it working. However, because we are bundling react we end up with a 'multiple instances of react' error that prevents users from actually using the bundle. The only ways I can see of fixing this are:
1. export react from the spectacle bundle so that Spectacle and the user's presentations are all made with the same instance of react.
2. keep react as external and alias it to `es-react` so that there is only one instance of react on the page (most likely via a rollup plugin - might need to be a custom plugin)
3. Use an import map, mapping react to `es-react`. 

What do you think of these options?

ryan-roemer:
Thanks for the summary. Let’s export react.

ryan-roemer:
@alex-saunders @kitten --  We've been changing direction for this PR for a bit, and I think it might help us to focus by splitting up the work we have here into more easily shippable and mergeable chunks:

- [ ] A UMD build with our rollup stuff here that supports `one-page` using what I've already done in https://github.com/FormidableLabs/spectacle/compare/task/esm-build...task/esm-build-one-page-tweaks . We will _not_ change `package.json:main|module`. This is just to support UMD `one-page`. Let's convert _this_ PR to do just that or split off a fresh branch/PR with just that first.
- [ ] Remove all `webpack` dependencies and switch `yarn start` workflow entirely over to rollup.

- ... after these are done, let's then consider the following potential tasks:
    - [ ] Remove and replace `babel` output with an `ESM` and `CJS` distribution and point `package.json:main|module`. We'll want consensus from the entire Spectacle rewrite team and @kale-stew specifically as our lead.
    - [ ] Consider an ESM dist consuming version of `one-page.html` that does not use any importmaps, shims, or the like. We don't necessarily need this right now and can just use UMD to ship rewrite first if needed.

alex-saunders:
@ryan-roemer sounds great. This work definitely needs to be split out.
I'll open a PR for the UMD build. I think it's probably easiest to cherry pick the rollup work out of this PR so that we can quickly iterate in a following PR to move the entire workflow over to rollup. 
As a side note: do we want to 'release' the rollup plugin as a separate package on the Formidable org? I've found it pretty useful for local dev matching up with the published unpkg module and ensuring that all dependencies have appropriate, browser-ready ESM builds.

alex-saunders:
Closing this PR in favour of separate Rollup build PRs which will include this work.

PR #732, opened 2019-10-07
Add UMD build config

alex-saunders:
- Adds UMD bundle config, generated by rollup
- Splits up config into self-contained 'factory' functions, within new 'config' directory'
- Adds `examples/one-page-umd.html` as an example of how to use the UMD bundle

TODO:
- ~replace user theme with global variable~*
- Adds docs on UMD example?

*I've added a basic example of a `window.userTheme` alternative to providing a custom theme. It's a little brittle as it requires users to explicity set the variable on `window` (or define it via `var`) and do so *before* the spectacle script is loaded. I'm not 100% certain this will be immediately obvious to users and could be prone to frustrating bugs?
Alternatively, we could have an optional `theme` prop on the top-level `Deck` component? This seems like a more intuitive (and react-y) solution but I'd be interested to hear people's opinion on this (I'm not sure if there would be opposition to this due to it being how the current version of spectacle handles custom themes - I think) cc @kale-stew @SunburtReynolds .


package.json
```diff
@@ -2,18 +2,19 @@
...
   "module": "es/index.js",
   "types": "index.d.ts",
   "scripts": {
-    "build": "",
     "lint": "eslint --ignore-path .gitignore .",
     "lint-fix": "",
     "prettier": "prettier \"**/*.{js,json,ts,css,md}\"",
     "prettier-fix": "npm run prettier -- --write",
     "deploy": "",
     "start": "builder run --env \"{\\\"BABEL_ENV\\\":\\\"esm\\\"}\" build-wds-base -- --port=3000 --hot",
```
ryan-roemer: If we're going full rollup and have buy-in from Spectacle team + @kale-stew let's go ahead and remove all `webpack` deps + build steps and replace `start` script here with a rollup version.

config/external.js
```diff
@@ -0,0 +1,21 @@
+const pkgInfo = require('../package.json');
+
+let external = ['dns', 'fs', 'path', 'url'];
```
ryan-roemer: Can we get code comments throughout all of this file to add more insight into (1) what we're doing and (2) why. Our target audience is your fellow maintainers -- folks experienced with at least webpack, but probably very unclear on what we're filtering on and why. Thanks!

.npmignore
```diff
@@ -1,7 +1,6 @@
 /*
 /dist/*
 !/dist/spectacle*
-/dist/*.map
```
ryan-roemer: In the past we didn't ship `*.map` files because they were _huge_ and people complained about package size. What's our current source map size and do we risk running into those problems again?

package.json
alex-saunders: Will do, I wasn't sure we wanted to remove + replace `webpack` in a future PR or within this piece of work. Happy to do it here.

.npmignore
alex-saunders: Our current map is pretty large (looking at around 700kB), I hadn't realised this was a problem and people had complained about package size. I'm happy to remove source maps as I suppose they're only at their most useful when developing Spectacle directly.

package.json
ryan-roemer: If it’s fast (like an hour or less), let’s do it here. If not separate pr. I’d like us to not have two builds for any longer than we need and I’d like the rewrite team to dogfood dev in rollup if there’s buy in immediately to flush out any dev issues early. Thanks!

ryan-roemer: If it’s fast (like an hour or less), let’s do it here. If not separate pr. I’d like us to not have two builds for any longer than we need and I’d like the rewrite team to dogfood dev in rollup if there’s buy in immediately to flush out any dev issues early. Thanks!

alex-saunders: I've had a crack at replacing webpack dev server with rollup plugins `serve` and `livereload` - accessible via `yarn start`.
I still need to clean up & comment the config a bit but it's probably in a fairly usable state now. Would be good to get some of the team trying out & testing the workflow.

config/plugins.js
```diff
@@ -0,0 +1,55 @@
...
+import serve from 'rollup-plugin-serve';
+import livereload from 'rollup-plugin-livereload';
+
+import babelPreset from './babel';
+
+const devServerPlugins = commandOptions => [
+  serve({
+    open: true,
+    contentBase: '.',
+    openPage: '/examples/one-page-umd.html',
```
ryan-roemer: As I mentioned in channel, dev server target is `ROOT/index.js` supporting either the mdx or js examples file imports. The current webpack html template is at `ROOT/index.html`.

ryan-roemer:
Can you catch this up with `task/rewrite` and harmonize `examples/one-page.html`? (Maybe just remove `one-page-umd.html`?

alex-saunders:
@ryan-roemer should be all ready now!

package.json
```diff
@@ -2,22 +2,20 @@
   "name": "spectacle",
   "version": "5.5.0",
   "description": "ReactJS Powered Presentation Framework",
+  "source": "src/components/index.js",
```
ryan-roemer: This should be `src/index.js`. That's what current "the root index" file is.

package.json
```diff
@@ -2,22 +2,20 @@
...
-    "build-wds": "webpack-dev-server --config webpack.config.example.js",
-    "build-webpack-base": "webpack",
-    "build-webpack-example": "builder run build-webpack-base -- --config webpack.config.example.js",
-    "build-webpack-lib-dev": "builder run build-webpack-base -- --config webpack.config.dev.js",
-    "build-webpack-lib": "builder run build-webpack-base",
-    "build": "builder concurrent --buffer build-webpack-lib build-webpack-lib-dev build-webpack-example",
     "lint": "eslint --ignore-path .gitignore .",
     "lint-fix": "yarn lint --fix",
     "prettier": "prettier \"**/*.{js,json,ts,css,md}\"",
-    "prettier-fix": "yarn prettier --write",
```
ryan-roemer: We use `yarn` for this project so let's standardize on that instead of `npm` for shelling commands.

package.json
```diff
@@ -2,22 +2,20 @@
...
+    "prettier-fix": "npm run prettier -- --write",
+    "rollup": "rollup -c config/rollup.config.js",
+    "dev-server": "yarn rollup --watch --open --port=3000",
+    "start": "NODE_ENV=development yarn dev-server",
+    "build": "NODE_ENV=production yarn rollup",
```
ryan-roemer: Let's make it cross-platform compatible with `builder` execution:

```js
"start": "builder run --env \"{\\\"NODE_ENV\\\":\\\"development\\\"}\" dev-server",
"build": "builder run --env \"{\\\"NODE_ENV\\\":\\\"production\\\"}\" rollup",
```

package.json
```diff
@@ -57,17 +61,23 @@
     "eslint-plugin-react": "^7.13.0",
     "eslint-plugin-react-hooks": "^1.6.0",
     "file-loader": "^4.2.0",
```
ryan-roemer: This is webpack right? Let's try reviewing and removing all of these.

package.json
```diff
@@ -57,17 +61,23 @@
...
-    "html-webpack-plugin": "^3.2.0",
     "prettier": "^1.15.3",
     "prop-types": "^15.7.2",
     "raw-loader": "^1.0.0",
```
ryan-roemer: Same here (?)

src/theme/index.js
```diff
@@ -1,6 +1,7 @@
 import defaultTheme from './default-theme';
 // see cli actions.js to understand this import
-let userTheme = require('spectacle-user-theme').default;
+// TODO: replace with global theme variable?
```
ryan-roemer: Let's add a ticket to support this comment and paste the ticket link in a code comment underneath this comment.

config/plugins/mdx-plugin/helpers.js
```diff
@@ -0,0 +1,110 @@
+import mdx from '@mdx-js/mdx';
```
ryan-roemer: We don't currently publish this (see: https://github.com/FormidableLabs/spectacle/blob/task/rewrite/.npmignore) -- is this intended to be just for _our_ examples only or are we planning on releasing our MDX support?

/cc @carlos-kelly @SunburtReynolds

config/rollup.config.js
```diff
@@ -0,0 +1,73 @@
...
+    ...config,
+    plugins: [
+      ...makePlugins(isProduction),
+      // if rollup has been ran with `--open`, include the dev server plugins
+      ...(commandOptions.open ? makeDevServerPlugins(commandOptions) : [])
+    ],
+    output: {
+      format: 'iife',
+      name: 'Spectacle',
+      file: path.join('dist', 'spectacle.js')
```
ryan-roemer: Can we name this `dist/example.js` or something so that it's very obvious what this is vs. the real library?

config/rollup.config.js
```diff
@@ -0,0 +1,73 @@
...
+  // UMD build
+  const umd = {
+    ...config,
+    external: ['react', 'react-is', 'react-dom', 'prop-types'],
+    plugins: [...makePlugins(isProduction)],
+    output: [
+      {
+        name: 'Spectacle',
+        sourcemap: false,
+        file: `./dist/umd/spectacle.js`,
```
ryan-roemer: This won't be published per https://github.com/FormidableLabs/spectacle/blob/task/rewrite/.npmignore (needs a `!` whitelist entry if we like this name).

Overall, I think you probably want to do a full build then `npm pack` and then view the tgz contents to make sure that everything you see in the tarball is correct. Also, we can remove `es|lib` directories as if we take this PR we won't produce babel transpiled files anymore.

examples/one-page.html
```diff
@@ -22,7 +22,7 @@
     <script src="https://unpkg.com/react-is@16.10.1/umd/react-is.production.min.js"></script>
     <script src="https://unpkg.com/prop-types@15.7.2/prop-types.min.js"></script>
     <!-- <script src="https://unpkg.com/spectacle/dist/spectacle.js"></script> -->
-    <script src="../dist/spectacle.js"></script>
+    <script src="../dist/umd/spectacle.js"></script>
```
ryan-roemer: This isn't working..

**Expected**: A full export of a `Spectacle` object that then hooks up our one-page inline JS/htm content.

**Actual**: `dist/umd/spectacle.js` appears to be built as an _application_ doing the `ROOT/index.js` demo (with MDX by default) that runs and shows a presentation. As soon as we get to our inline script tage, it errors out with `Uncaught ReferenceError: Spectacle is not defined` because presumably it's not an actual UMD exported lib.

package.json
```diff
@@ -2,22 +2,20 @@
   "name": "spectacle",
   "version": "5.5.0",
   "description": "ReactJS Powered Presentation Framework",
+  "source": "src/components/index.js",
   "main": "lib/index.js",
   "module": "es/index.js",
```
ryan-roemer: `lib/` and `es/` won't exist with this infrastructure. Let's make sure that we point these to workable, appropriate CJS and ESM rollup bundles.

package.json
kitten: Quick suggestion, we could use cross-env here which is a little easier to read and more commonly used?

ryan-roemer: For concurrent/parallel work, `builder concurrent` (run different `package.json:scripts` tasks) or `builder env` (run the same `package.json:scripts` task with different environment variables) is more flexible and powerful than `concurrently` and is also part of `builder`.

If we're additionally doing that, then I'd vote we stick with `builder` to just let one tool do everything. If not, then sure, go ahead with `cross-env`

On that note now, are we producing both minified and non-minified ("development") builds in `dist/` for the library? If not, we should consider doing that. (I mention it because what `builder` was here in the first place for last time...)

ryan-roemer:
As a quick follow on task, let's maybe create a quick repository that then does a webpack build of (1) a normal deck and (2) an MDX deck to make sure we can still support our user base who wants to be on those build tools.

(Or we could do it in `examples`, but it should consume `dist/**/*.js` files and not `src/**/*.js` files)

config/rollup.config.js
alex-saunders: I was intending to have the different build formats seperated out by directory within the `dist` directory - i.e. `dist/es/spectacle.js`, `/dist/cjs/spectacle.js`, `dist/umd/spectacle` (and the relevant fields in `package.json` pointing to these locations) - should we remove `dist` from the npmignore (but keep ignoring source maps)?

examples/one-page.html
alex-saunders: My bad, needed to change the source location to `components/index.js` rather than `~/index.js`.
I'm inclined to keep the `package.json` `source` field as `components/index.js` to reflect this as the 'library entry' rather than the dev-server entry, what's your opinion on this?

package.json
alex-saunders: I'll remove these entries for now as we'll work on these bundles in the next PR.

config/babel.js
```diff
@@ -1,5 +1,6 @@
 const babelPreset = isProduction => ({
   babelrc: false,
+  compact: false,
```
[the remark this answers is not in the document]
alex-saunders: I thought `compact` was babel-speak for minify but looks like there's a separate `minified` option I should use so I'll remove this and add minified instead.

alex-saunders:
@ryan-roemer should we do this after we have created the CJS bundle as this is more likely to be what users will end up consuming via webpack rather than the UMD module that is created via this PR?

ryan-roemer:
@alex-saunders -- Good point about CJS bundles. Let's make sure we output:

- UMD dist
- Something that works for CJS `package.json:main`
- Optionally, something that works for ESM `package.json:module`

alex-saunders:
Will do, I'll create a quick PR after this has landed to add the CJS bundle as that should be fairly simple with this setup.
Then we'll get working on the ESM output.

config/rollup.config.js
ryan-roemer: You can adjust the `.npmignore` to match what currently happens on `master`, namely:

1. Ignore the example bundle for the demo app
2. Include the production and development consumable entry points (UMD, CJS, ESM, etc.)

examples/one-page.html
ryan-roemer: `src/components/index.js` doesn't even exist anymore. We've removed it and moved **everything** to https://github.com/FormidableLabs/spectacle/blob/task/rewrite/src/index.js as we're shipping more than components in our final deliverable bundle.

config/rollup.config.js
```diff
@@ -0,0 +1,78 @@
...
+import makePlugins from './plugins/index';
+import makeDevServerPlugins from './plugins/server';
+
+const pkgInfo = require('../package.json');
+
+const isProduction = process.env.NODE_ENV === 'production';
+
+// default config that is used across all builds
+const config = {
+  input: pkgInfo.source || './index.js',
```
ryan-roemer: `./index.js` is the example deck, so I'm not sure this is a good default (although for a different project would totally make sense). Maybe point to `./src/index.js`?

config/rollup.config.js
```diff
@@ -0,0 +1,78 @@
...
+    // from illegal property access.
+    propertyReadSideEffects: false
+  }
+};
+
+export default function makeConfig(commandOptions) {
+  const builds = [];
+
+  // iife build (for use with dev-server)
+  const iife = {
```
ryan-roemer: Let's maybe rename this variable to `example` so it's clear it's only the example deck (which we don't publish to npm).

package.json
```diff
@@ -42,32 +39,41 @@
...
     "@babel/preset-env": "^7.4.5",
     "@babel/preset-react": "^7.0.0",
     "@mdx-js/loader": "^1.0.0-rc.4",
     "babel-eslint": "^10.0.2",
-    "babel-loader": "^8.0.6",
+    "babel-plugin-module-resolver": "^3.2.0",
+    "babel-plugin-transform-async-to-promises": "^0.8.14",
+    "babel-plugin-transform-react-remove-prop-types": "^0.4.24",
     "builder": "^4.0.0",
+    "es-react": "^16.8.60",
```
ryan-roemer: I don't think we're using this anywhere in this PR.

alex-saunders:
Closing this in favour of renaming branch to `task/rewrite-rollup`

PR #735, opened 2019-10-15
Add one-page with UMD build.

ryan-roemer:
This is intended to tide us over until we figure out the ESM + rollup + everything else infrastructure discussion. We can rip all this out later but might be a good move for speed right now.

- Adds actual `dist/*` builds for `example`, `spectacle` (lib prod), `spectacle.dev`
- Adds `examples/one-page.html`
- Adds some TODO code comments
- Misc `package.json:scripts` cleanup and TODO additions
- Makes ESM the default for `.babelrc` and instead has new `BABEL_ENV=cjs` for CommonJS since that is now the one-off, less used use case.

/cc @carlos-kelly @alex-saunders


examples/one-page.html
```diff
@@ -0,0 +1,170 @@
...
+              gridTemplateRows="1fr 1fr 1fr"
+              alignItems="center"
+              justifyContent="center"
+              gridRowGap=${15}
+            >
+              ${Array(9)
+                .fill('')
+                .map(
+                  (_, index) =>
+                    // TODO(ONE-PAGE): I don't see anything in dev or one-page (???)
```
carlos-kelly: What is this comment referring to?

index.js
```diff
@@ -1,6 +1,9 @@
 import React from 'react';
 import { render } from 'react-dom';
 
+// TODO: This should be moved to `examples` along with `index.html` and not
+// be in real repo root (unless this is our "cloned repo" example??)
```
carlos-kelly: This currently is a development example and will be extracted before release. Good catch!

src/index.js
```diff
@@ -0,0 +1,46 @@
+// TODO: Consider removing `./components/index` here and just doing right here.
+// TODO: Consider additional re-exports from other things.
```
carlos-kelly: Good idea!

PR #736, opened 2019-10-17
Fix CLI build process; Add title/port options.

carlos-kelly:
* Fix the CLI; currently it shows an empty screen since the base webpack options were shuffled around a bit
* Add title and port options to the CLI

Screen Shot 2019-10-17 at 12 54 46 PM
Screen Shot 2019-10-17 at 12 54 30 PM


src/components/image.js
```diff
@@ -5,11 +5,14 @@ const Image = styled('img')(
...
-  `
+  )
 );
 
-export { Image };
+const FullSizeImage = styled(Image);
+
+FullSizeImage.defaultProps = {
+  maxWidth: '100%',
+  maxHeight: '100%'
```
carlos-kelly: These props broke regular sized images, but fit the requirement for the full sized ones. Break it out into two components since a one-size-fits-all didn't work.

src/index.js
```diff
@@ -1,27 +1,22 @@
-// TODO: Consider removing `./components/index` here and just doing right here.
-// TODO: Consider additional re-exports from other things.
-
+import Deck from './components/deck';
+import Slide from './components/slide';
+import Appear from './components/appear';
+import CodePane from './components/code-pane';
```
carlos-kelly: This is just addressing Roemer's comments about the double-index. Unifies them.

bin/actions.js
```diff
@@ -21,12 +20,14 @@ const launchServer = (configUpdates = {}) => {
...
 };
 
-const launchMDXServer = (mdxFilePath, themeFilePath) => {
+const launchMDXServer = (mdxFilePath, themeFilePath, title, port = 3000) => {
   if (!mdxFilePath) {
     // developer error - must supply an entry file path
     throw new Error('MDX file path must be provided.');
   }
 
+  console.log(title);
```
SunburtReynolds: Probs want to remove this.

PR #740, opened 2019-10-23
Add base JS-based auto sizing

carlos-kelly:
The idea is to allow for CodePane to consume the empty space around the footer and elements automatically with an opt-in prop. Prior, the user would have to either calculate or best-guess the max-height value to do so. This applies for CodePanes with lots of content. This full-height calculation can be extended to other components down the road.

Screen Shot 2019-10-25 at 12 17 28 PM

### DIFF
--- a/examples/js/test-js.js
+++ b/examples/js/test-js.js
@@ -26,13 +26,13 @@ const cppCodeBlock = `#include <iostream>
 #include <sstream>
 #include <pthread.h>
 
-struct thread_data_t 
+struct thread_data_t
 {
    int  thread_id;
    std::string message;
 };
 
-void *print_thread_message(void *thread_arg) 
+void *print_thread_message(void *thread_arg)
 {
    struct thread_data_t *thread_data;
    thread_data = (struct thread_data_t *) thread_arg;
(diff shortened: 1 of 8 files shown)


examples/js/test-js.js
```diff
@@ -22,13 +22,48 @@ import {
...
+    message << "The math result is " << answer << "!";
+    thread_data.thread_id = i;
+    thread_data.message = message.str();
+    int err = pthread_create(&threads, NULL, print_thread_message, (void *)&thread_data[i]);
+    
+    if (err) 
+    {
+      exit(-1)
+    }
+  }
```
carlos-kelly: This is all demo C++ to make the code pane bigger. No need to review this.

ryan-roemer: Are you sure this is thread-safe? :trollface:

stevenmusumeche: 💅

src/hooks/use-autofill-height.js
```diff
@@ -0,0 +1,77 @@
+import * as React from 'react';
```
stevenmusumeche: very cool

PR #741, opened 2019-10-24
Replace webpack with rollup

alex-saunders:
This PR contains all the work for replacing webpack with rollup. It's based off [this initial PR](https://github.com/FormidableLabs/spectacle/pull/732).

Current TODO list is:
- ~Generate UMD build~
- Generate CJS build (#742)
- Generate ESM build
- ~Replace webpack-dev-server with rollup equivalent dev server~
- Replace the CLI webpack functionality with rollup, ensuring custom user themes remain usable


config/external.js
```diff
@@ -0,0 +1,24 @@
+const pkgInfo = require('../package.json');
+
+let external = ['dns', 'fs', 'path', 'url'];
+
+// add all dependencies defined in package.json into external array
+// as we don't want to bundle these
+if (pkgInfo.peerDependencies)
+  external.push(...Object.keys(pkgInfo.peerDependencies));
+if (pkgInfo.dependencies) external.push(...Object.keys(pkgInfo.dependencies));
```
ryan-roemer: Do we need `optionalDependencies` too? These are generally considered dependencies if available.

config/rollup.config.js
```diff
@@ -0,0 +1,78 @@
...
+  // if rollup has been ran via `--open` then only
+  // include the dev-server IIFE build.
+  if (commandOptions.open) {
+    builds.push(...[example]);
+  }
+
+  // if we are not running the dev-server, include all bundle builds
+  // that we publish to npm (including production & development).
+  if (!commandOptions.open) {
+    builds.push(...[makeUmd(true), makeUmd(false)]);
```
ryan-roemer: We still have `NODE_ENV=development|production` in our scripts, yet it seems to be that these aren't used.

Separately we're doing serial builds right whereas if we did single builds based on envvars like already exists for webpack in `task/rewrite` we could be doing parallel with `builder` (or `concurrently` or whatever).

package.json
```diff
@@ -2,28 +2,25 @@
...
     "lint-fix": "yarn lint --fix",
     "prettier": "prettier \"**/*.{js,json,ts,css,md}\"",
-    "prettier-fix": "yarn prettier --write",
-    "deploy": "echo TODO_IMPLEMENT_OR_REMOVE",
-    "start": "builder run build-wds -- --port=3000 --hot",
+    "prettier-fix": "yarn prettier -- --write",
+    "rollup": "rollup -c config/rollup.config.js",
+    "dev-server": "yarn rollup --watch --open --port=3000",
+    "start": "builder run --env \"{\\\"NODE_ENV\\\":\\\"development\\\"}\" dev-server",
+    "build": "builder run --env \"{\\\"NODE_ENV\\\":\\\"production\\\"}\" rollup",
```
ryan-roemer: These `NODE_ENV` vars don't appear to be used at all.

package.json
```diff
@@ -2,28 +2,25 @@
...
   "version": "5.5.0",
   "description": "ReactJS Powered Presentation Framework",
-  "main": "lib/index.js",
-  "module": "es/index.js",
+  "source": "src/index.js",
+  "main": "dist/umd/spectacle.js",
   "types": "index.d.ts",
   "scripts": {
-    "clean": "rimraf dist lib es",
+    "clean": "rimraf dist",
```
ryan-roemer: We shouldn't remove `lib es` from clean until our CJS and ESM replacement PRs land.

package.json
```diff
@@ -62,19 +63,23 @@
...
+    "rollup-plugin-commonjs": "^10.1.0",
+    "rollup-plugin-image": "^1.0.2",
+    "rollup-plugin-livereload": "^1.0.4",
+    "rollup-plugin-node-resolve": "^5.2.0",
+    "rollup-plugin-replace": "^2.2.0",
+    "rollup-plugin-serve": "^1.0.1",
+    "rollup-plugin-terser": "^5.1.2",
     "tslint": "^5.12.1",
-    "typescript": "^3.2.2",
-    "webpack": "^4.40.2",
```
ryan-roemer: It looks like we haven't removed the `ROOT/webpack*.js` files yet in this PR?

config/external.js
alex-saunders: Sure, seems a reasonable thing to add.

config/rollup.config.js
alex-saunders: Ahh good idea, I'll revert this to env variables and go with a parallel build approach.

package.json
alex-saunders: I was going to do that once all build format PRs are in but I'm happy to do it now

config/external.js
```diff
@@ -0,0 +1,24 @@
...
+
+// optionally remove some dependencies from the external that don't have
```
ryan-roemer: Per my top level comment, let's remove this.

config/plugins/index.js
```diff
@@ -0,0 +1,63 @@
...
+    // resolve modules located in node_modules
+    nodeResolve({
+      // first look for `browser` and `module` fields in
+      // module package.json to find ESM builds
+      mainFields: ['module', 'jsnext', 'main'],
+      browser: true
+    }),
+    // allow images to be directly imported á la webpack
+    image(),
+    // smooth over commonjs exports, converting them to an ES compatible version
```
ryan-roemer: Let's only keep what we need for the UMD outputs of these configurations.

config/plugins/mdx-plugin/index.js
```diff
@@ -0,0 +1,20 @@
+import { extname } from 'path';
```
ryan-roemer: Do we want to ship our rollup mdx plugin (because right now the mdx loader is not available as npm published)? We already publish `/mdx-loader` for webpack. Perhaps we could ticket out converting that to `/mdx-loader/{webpack,rollup}` and publishing both from the same directory so folks can create Spectacle decks with either...

package.json
```diff
@@ -2,8 +2,8 @@
   "name": "spectacle",
   "version": "5.5.0",
   "description": "ReactJS Powered Presentation Framework",
-  "main": "lib/index.js",
-  "module": "es/index.js",
```
ryan-roemer: There is still `es/` babel output in this PR. Let's leave it until its time to remove.

package.json
```diff
@@ -2,8 +2,8 @@
...
+  "source": "src/index.js",
+  "main": "dist/umd/spectacle.js",
```
ryan-roemer: This should remain unchanged. `umd` is not our CJS entry point.

config/plugins/mdx-plugin/index.js
alex-saunders: That sounds like a great idea, the rollup plugin needs a little cleaning up (and sourcemaps need to be configured) before it's available for release but that's definitely something we should work on.

ryan-roemer:
This has been a fantastic POC/experiment, but it seems we've got everything shipping in good form with spectacle, spectacle-cli, spectacle-boilerplate, and spectacle-mdx-loader on webpack solidly. The main impetus of one-page.html is getting by in modernized htm form but without true ESM modules (but given our limitations there because of upstream deps, I think we're not going to be there any time soon).

... so closing, but perhaps we should ticket out rollup support / investigation in the future!

PR #746, opened 2019-10-28
Table component branch bug

kiraarghy:
Markdown example to demonstratte indentation problem with prettier and marksy.

In this [case](https://github.com/FormidableLabs/spectacle/pull/746/files#diff-27d1398137d07ae13f238fccada6a73fR90-R101) marksy interprets the code as a codepane, not a table. After having it explained by Carlos, this makes sense but this is probably a sub-optimal user experience and we should think about how to solve this.

### DIFF
--- a/examples/js/test-js.js
+++ b/examples/js/test-js.js
@@ -87,18 +87,6 @@ const TestJs = () => (
     )}
   >
     <Slide>
-      <Slide>
-        <Markdown>{`
-        # hello there
-        
-        | hi   | Supported | Versions |
-|-----------|-----------|----------|
-| Chrome          | Yes       | Last 2   |
-| Firefox         | Yes       | Last 2   |
-| Opera           | Yes       | Last 2   |
-| Edge (EdgeHTML) | No        |          |
-| IE 11           | No        |          |`}</Markdown>
-      </Slide>
       <Heading>Spectacle</Heading>
       <Text>Hello There 🤗</Text>
       <Quote>This is a Formidaquote!</Quote>
@@ -184,6 +172,21 @@ const TestJs = () => (
             </FlexBox>
           ))}
       </Grid>
+    </Slide>
+    <Slide>
+      <Markdown>
+        {`
+          # hello there
+
+          | hi              | Supported | Versions |
+          |-----------------|-----------|----------|
+          | Chrome          | Yes       | Last 2   |
+          | Firefox         | Yes       | Last 2   |
+          | Opera           | Yes       | Last 2   |
+          | Edge (EdgeHTML) | No        |          |
+          | IE 11           | No        |          |
+        `}
+      </Markdown>
     </Slide>
     <Slide>
       <Markdown>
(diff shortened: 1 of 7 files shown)


kitten:
It’s be great to just apply a simple strip-indent to the content of the markdown code, similar to: https://github.com/sindresorhus/strip-indent

examples/js/test-js.js
```diff
@@ -87,6 +87,18 @@ const TestJs = () => (
     )}
   >
     <Slide>
+      <Slide>
+        <Markdown>{`
+        # hello there
+        
+        | hi   | Supported | Versions |
+|-----------|-----------|----------|
```
carlos-kelly: This indentation looks strange.

kiraarghy:
Oh I have a separate branch that I was going to have a look at the styling in, will get that up soon!

kiraarghy:
> It’s be great to just apply a simple strip-indent to the content of the markdown code, similar to: https://github.com/sindresorhus/strip-indent

I'm not sure if there is a simple solution to this as @carlos-kelly and @ryan-roemer have said if I want to put spacing in my markdown if we strip out all indent it would screw up that spacing? 

I'm not sure if this would solve the issue highlighted [here](https://github.com/FormidableLabs/spectacle/pull/746/files#diff-27d1398137d07ae13f238fccada6a73fR90-R101) as the least number of leading whitespace is line 95?

examples/js/test-js.js
kiraarghy: Yeah it's wrong, it's just not immediately noticeable that it's wrong 😭

ryan-roemer:
As an FYI `strip-indent` is exactly the same in function as our indent normalization function, it just uses two packages (`min-indent` and `strip-indent`). We're likely going to need something different.

examples/js/test-js.js
ryan-roemer: For this specific case, I don't see how we're going to handle it outside of something super complicated like probablistically calculating indent based on markdown parsing.

The only idea here I can come up with is "indent cannot be `0`" and ignore a detected min indent that is `0` relying on everything else....

ryan-roemer:
The original example had incorrect `<Slide>` nesting that threw everything off. I've:

- Fixed the nesting
- Moved the slide to the markdown examples section with others
- Updated `.prettierignore` with our current outputs
- Added a comparable `one-page.html` table slide

With this commit, it appears that just doing something like:

```js
    <Slide>
      <Markdown>
        {`
          # hello there

          | hi              | Supported | Versions |
          |-----------------|-----------|----------|
          | Chrome          | Yes       | Last 2   |
          | Firefox         | Yes       | Last 2   |
          | Opera           | Yes       | Last 2   |
          | Edge (EdgeHTML) | No        |          |
          | IE 11           | No        |          |
        `}
      </Markdown>
    </Slide>
```

Works fine. We still need to do the nested `{\`STRING\`}` thing, but maybe that's not too bad?

And, I've run everything through `yarn prettier-fix` and I don't see prettier doing any further mutations or changes to the table markdown string.

kiraarghy:
@ryan-roemer that's amazing, thanks!

PR #748, opened 2019-10-29
Adds more styled-system options to table and move table to table.js

kiraarghy:
Adds more styled-system options to table.

Moves tabel from `/typography.js` to `/table.js` for clarity.


carlos-kelly:
:shipit:

PR #749, opened 2019-10-29
Add ESM Rollup Build

alex-saunders:
This PR:
- Adds the rollup config for the ESM build
- Adds the `examples/one-page-esm` example project showing how to use the ESM bundle
	- (There's tiny server file required to run this project that proxies the request for spectacle to the correct folder as we cannot import from above the root folder with ES modules)
- Adds the custom unpkg rollup plugin to allow for local development (replaces all imports with UNPKG `?module` URLs)
	- Maybe we want to split this out into its own project?
- Updates package.json scripts to allow specific formats to be built, i.e. you could now run `yarn build:prod --format=esm,cjs` to just product production, CJS and ESM builds.

The branch is originally based off `task/rewrite-rollup-cjs` to include the work I have done there, but I will change the base branch once that is merged.


config/plugins/unpkg-plugin/index.js
```diff
@@ -0,0 +1,103 @@
+const fs = require('fs');
+const { walk } = require('estree-walker');
+const MagicString = require('magic-string');
+const readPkg = require('read-pkg');
+
+const pkgInfo = readPkg.sync();
```
kitten: I think this would be a great small package!

config/rollup.config.js
```diff
@@ -57,6 +59,21 @@ export default function makeConfig(commandOptions) {
         sourcemap: false
       }
     ]
+  });
+
+  const makeESM = () => ({
+    ...config,
+    // These packages do not have a browser-ready ESM build, so we'll just
+    // bundle and ship them
+    external: external.filter(
```
kitten: Word of caution here again; we don’t want to point `package.json:module` at this if we internalise styled-components.

If someone ends up using Spectacle with CRA and/or custom components that use styled-components, then we’ll have it twice in the bundle and it’ll likely break

alex-saunders: @kitten would you suggest making an ESM version with no external deps bundled  (that wouldn't work directly in the browser) and pointing `package.json:module` at that? Then we could create a separate 'browser' bundle with styled-components bundled and point `package.json:browser` at that?

PR #757, opened 2019-10-31
Remove spectacle-user-theme webpack alias strategy.

ryan-roemer:
As discussed internally, we don't need to webpack alias a made up `spectacle-user-theme` thing to allow user-provided theme overrides via manual `Deck` or the CLI. We can just pass them as props to the top level exported deck and much more simply have it all work out.

Work:

- Removed unneeded files and webpack alias.
- Add override example in `one-page.html`

Example:

```js
<Deck
    theme={{
      colors: {
        secondary: 'yellow'
      }
    }}
  >
    <Slide>
```

To test out:

```sh
$ git checkout chore/remove-theme-alias
$ yarn build
$ open examples/one-page.html
```

You should see a pink heading (colored `secondary`) on the first page instead of default theme color. Try out other theme overrides in `examples/one-page.html` and refresh to see in action!

### DIFF
--- a/examples/one-page.html
+++ b/examples/one-page.html
@@ -9,14 +9,6 @@
   <body>
     <div id="root"></div>
 
-    <script>
-      window.userTheme = {
-        colors: {
-          secondary: 'hotpink'
-        }
-      };
-    </script>
-
     <script src="https://unpkg.com/react@16.10.1/umd/react.production.min.js"></script>
     <script src="https://unpkg.com/react-dom@16.10.1/umd/react-dom.production.min.js"></script>
     <script src="https://unpkg.com/react-is@16.10.1/umd/react-is.production.min.js"></script>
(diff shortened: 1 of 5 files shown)


examples/one-page.html
```diff
@@ -9,14 +9,6 @@
   <body>
     <div id="root"></div>
 
-    <script>
-      window.userTheme = {
```
ryan-roemer: Note: This never worked. I took it from a different branch that had integration.

src/components/deck/index.js
```diff
@@ -273,7 +273,7 @@ Deck.propTypes = {
 };
 
 const ConnectedDeck = props => (
-  <ThemeProvider theme={theme}>
+  <ThemeProvider theme={mergeTheme(props.theme)}>
     <TransitionPipeProvider>
       <Deck {...props} />
```
ryan-roemer: One thing I'm not sure on -- do I need to pass the merged `theme` here too? So like the equivalent of:

```js
<Deck {...props, ...{ theme: mergeTheme(props.theme) }} />
```

(But I'd do only one call to `mergeTheme()` in actual code).

PR #758, opened 2019-10-31
Kara/print and export

kiraarghy:
This ticket implements the export and print functionality that was in original Spectacle.

When a user uses the `exportMode` flag, when they print it will output each slide as a single page.
exportmode

When a user uses the `printMode` flag, it will apply a print theme and a white background, saving on printing. 

The printTheme has alright contrast at the moment, though am more than happy to take input on it @alex-saunders you're pretty good at this kinda thing. 

I have also added a lightTheme for the codepane when in `printMode`.

greyexport

### DIFF
--- a/src/components/slide.js
+++ b/src/components/slide.js
@@ -94,6 +94,17 @@ const Slide = props => {
     };
   }, [transformForWindowSize, scaleRatio]);
 
+  const transforms = React.useMemo(
+    () =>
+      state.exportMode
+        ? {}
+        : {
+            transform: `scale(${ratio})`,
+            transformOrigin: `${origin.x} ${origin.y}`
+          },
+    [state.exportMode, origin, ratio]
+  );
+
   const value = useSlide(slideNum);
   const { numberOfSlides } = value.state;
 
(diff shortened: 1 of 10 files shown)


src/components/slide.js
```diff
@@ -95,11 +102,17 @@ const Slide = props => {
...
+      backgroundColor={
+        window.location.search.includes('print') ? '#ffffff' : backgroundColor
+      }
+      style={
+        state.exportMode
+          ? {}
+          : {
+              transform: `scale(${ratio})`,
+              transformOrigin: `${origin.x} ${origin.y}`
+            }
```
carlos-kelly: This is starting to get a little big, perhaps time to move to a React.useMemo?

src/components/slide.js
```diff
@@ -95,11 +102,17 @@ const Slide = props => {
   return (
     <SlideContainer
       ref={slideRef}
-      backgroundColor={backgroundColor}
-      style={{
-        transform: `scale(${ratio})`,
-        transformOrigin: `${origin.x} ${origin.y}`
-      }}
+      backgroundColor={
+        window.location.search.includes('print') ? '#ffffff' : backgroundColor
```
carlos-kelly: Could we get this from the `useUrlRouting` hook?

src/theme/index.js
```diff
@@ -1,14 +1,33 @@
...
 
 // Merge a user-provided theme in with defaults.
 // **Note**: Assumes theme objects only go 2 levels deep.
-export const mergeTheme = theme =>
-  Object.keys(theme || {}).reduce((merged, key) => {
-    merged[key] = { ...merged[key], ...theme[key] };
-    return merged;
-  }, deepCopy(defaultTheme));
-
+export const mergeTheme = theme => {
```
ryan-roemer: @kiraarghy and I slack chatted a minor refactor here.

src/components/slide.js
kiraarghy: Sure can do!

src/components/slide.js
kiraarghy: Missed this, will get this done!

PR #759, opened 2019-10-31
Added support for flex-based AutoLayout

carlos-kelly:
The goal is to enable a quick way to enable flex to allow for content to be center positioned in the slide and flex column. This should allow for quicker layout when not wanting to use Box/Flex/Grid components or when using MDX.

This prop does two things

1. Makes the template and slide content flex siblings since you cannot control the template overlapping in MDX mode
2. Center-positions the slide content vertically using flex

Screen Shot 2019-10-31 at 2 12 10 PM

Screen Shot 2019-10-31 at 2 17 23 PM

Example showing the layout margins of the flex siblings of slide content and template:
Screen Shot 2019-10-31 at 2 16 55 PM
Screen Shot 2019-10-31 at 2 17 10 PM

### DIFF
--- a/src/components/deck/index.js
+++ b/src/components/deck/index.js
@@ -95,13 +95,17 @@ const mapMarkdownIntoSlides = (child, index) => {
   return child;
 };
 
-const Deck = ({
-  children,
-  loop,
-  keyboardControls,
-  animationsWhenGoingBack,
-  ...rest
-}) => {
+const Deck = props => {
+  const {
+    children,
+    loop,
+    keyboardControls,
+    animationsWhenGoingBack,
+    autoLayout,
+    backgroundColor,
+    textColor,
+    template
+  } = props;
   if (React.Children.count(children) === 0) {
     throw new Error('Spectacle must have at least one slide to run.');
   }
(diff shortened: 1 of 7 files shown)


src/components/deck/index.js
```diff
@@ -86,13 +86,17 @@ const mapMarkdownIntoSlides = (child, index) => {
...
+  const {
+    children,
+    loop,
+    keyboardControls,
+    animationsWhenGoingBack,
+    autoLayout,
+    backgroundColor,
+    textColor,
+    template
+  } = props;
```
carlos-kelly: I'm updating this to be explicit about the props, no more `...rest`.

src/components/slide.js
```diff
@@ -10,20 +10,51 @@ const SlideContainer = styled('div')`
   width: ${({ theme }) => theme.size.width || 1366}px;
   height: ${({ theme }) => theme.size.height || 768}px;
   overflow: hidden;
+  ${({ autoLayout }) =>
+    autoLayout &&
+    css`
+      height: 100%;
+      display: flex;
+      flex-direction: column-reverse;
```
carlos-kelly: We use `column-reverse` because the template is higher up in the DOM order because in non-auto layout mode it's an overlapping div and we don't want it to obstruct slide content.

src/hooks/use-autofill-height.js
```diff
@@ -8,14 +8,20 @@ const getNodeFullHeight = node => {
...
-  );
+    parseFloat(style.marginBottom);
+  return height;
 };
 
+const isCurrentNodeAutoFill = current =>
+  current.classList.contains('spectacle-auto-height-fill') ||
+  (current.tagName.toLowerCase().includes('pre') &&
+    current.childNodes &&
+    current.childNodes[0].classList.contains('spectacle-auto-height-fill'));
```
carlos-kelly: mdx/react adds a second `pre` tag as a child, so we need to account for that when allowing for auto-height components.

kale-stew:
I'll add documentation surrounding this feature in an upcoming docs PR so this doesn't get held up any longer. Surprisingly few merge conflicts rebasing behind the latest! Exciting stuff

PR #760, opened 2019-11-01
keyboardControls

kiraarghy:
This branch provides feature parity with existing spectacle for the keybindings for setting specific modes. 

Alt + P -> Presenter mode.
Alt + O -> Overview mode.
Alt + F -> Fullscreen mode. 

https://github.com/FormidableLabs/spectacle#controls

### DIFF
--- a/src/hooks/use-keyboard-controls.js
+++ b/src/hooks/use-keyboard-controls.js
@@ -1,12 +1,15 @@
 import React from 'react';
 import debounce from '../utils/debounce';
+import { useToggleFullScreen } from './use-full-screen';
 
 const useKeyboardControls = ({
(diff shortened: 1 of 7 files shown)


kiraarghy:
FWIW I do not have a windows computer to test on so I assume these work.

src/hooks/use-keyboard-controls.js
```diff
@@ -33,14 +36,35 @@ const useKeyboardControls = ({
             e.preventDefault();
           }
         }
+        if (!!e.altKey) {
```
kiraarghy: I guess there may be an argument that this hook is doing too much. 

I like that all the logic is co-located but happy for feedback.

PR #762, opened 2019-11-02
Support Notes for MD parsing

carlos-kelly:
Support Notes for the Markdown Slides Component

```jsx
<Markdown containsSlides>
      {`
        ### First MD generated Slide
        Notes: These are notes
        ---
        ### Second MD Generated Slide
        Notes: These are more notes
      `}
</Markdown>
```

Screen Shot 2019-11-02 at 12 36 19 PM
Screen Shot 2019-11-02 at 12 36 25 PM

### DIFF
--- a/src/utils/notes.js
+++ b/src/utils/notes.js
@@ -1,0 +1,27 @@
+const NOTES_MARKER = 'Notes: ';
+const NOTES_REG = new RegExp(`^${NOTES_MARKER}`, 'm');
+/*
+ * We want to pull the notes out of each slide. This RegEx looks for "Notes: "
+ * that starts on a new line. Anything after the notes marker will be considered notes.
+ */
+export const isolateNotes = content => {
+  const indexOfNotes = content.search(NOTES_REG);
+  if (indexOfNotes >= 0) {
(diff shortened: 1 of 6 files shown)


src/utils/notes.js
```diff
@@ -0,0 +1,27 @@
...
+ * compiled by mdx.sync. Remove all the notes content.
+ */
+export const removeNotes = content => {
+  const indexOfNotes = content.search(NOTES_REG);
+  if (indexOfNotes >= 0) {
+    // found the notes marker!
+    return content.substring(0, indexOfNotes);
+  }
+  return content;
+};
```
carlos-kelly: These functions are from the `spectacle-cli` package. Next update to the CLI will use these instead of having them in that repo.

src/utils/notes.js
```diff
@@ -0,0 +1,27 @@
+const NOTES_MARKER = 'Notes: ';
```
rgerstenberger: One thought is we may want this to be dynamic from the CLI or some config later. I'm not sure it's possible to override with this structure. Or we can say this is the API for md(x) notes and it's static :)

PR #767, opened 2019-11-04
Task/testing

kiraarghy:
Initialises jest config and a couple of tests for deck components.


src/hooks/use-url-routing.test.js
```diff
@@ -0,0 +1,308 @@
+import React from 'react';
+import Enzyme, { mount } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import * as history from 'history';
```
kiraarghy: Unsure how to test the `stateFromUrl` and `onHistoryChange` functions in `use-url-routing` but the external APIs are tested.

PR #804, opened 2020-01-17
Task/rewrite docs props

carlos-kelly:
Add all the composable styled system props and what tags use them


kale-stew:
😍 Thank you

PR #807, opened 2020-01-21
Use sane defaults for layout and eliminate autoLayout prop.

carlos-kelly:
Establish sane defaults for layout.

**Default style**
Screen Shot 2020-01-21 at 10 11 34 AM

**Optional center-center style** using `<FlexBox height="100%" flexDirection="column">`
Screen Shot 2020-01-21 at 10 11 25 AM


src/components/layout.js
```diff
@@ -33,13 +22,7 @@ FlexBox.defaultProps = {
...
 };
 
-const Grid = styled('div')(
-  compose(
-    layout,
-    grid,
-    position
-  )
-);
+const Grid = styled('div')(compose(layout, grid, position));
```
kale-stew: Looks like some prettier changes are being reversed, did you run `yarn format`?

carlos-kelly: I did! Prettier actually changed it to the new diff.

kale-stew: i ask because this has been the common cause of failure for me in CI

```
Checking formatting...
src/components/image.js
src/components/table.js
src/components/typography.js
Code style issues found in the above file(s). Forgot to run Prettier?
```

carlos-kelly: We may need to define a column size in prettier config? I'll look into it.

PR #827, opened 2020-02-03
Added Theme docs, fix default margins

carlos-kelly:
Description

- Adds documentation on the theming system
- Removes fixed values for margins for base tags. Replaces them with scale-based sizes in default theme.

### DIFF
--- a/docs/content/themes.md
+++ b/docs/content/themes.md
@@ -1,0 +1,101 @@
+---
+title: Theme System
+order: 6
+---
+
+<a name="theme-system"></a>
+
+# Theme System
+
+Spectacle has a robust theme system that is built upon [styled system](https://styled-system.com/theme-specification).
+
+A theme is a 2-level deep object of theme keys with label and CSS property object values.
+
+### Simple Theme Object Example
+
+```js
+const theme = {
+  colors: {
+    primary: '#f00',
+    secondary: '#00f'
+  },
+  fontSizes: {
+    header: '64px',
+    paragraph: '28px'
+  }
+};
+```
+
+### Usage Example
+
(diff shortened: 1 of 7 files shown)


src/components/typography.js
```diff
@@ -8,7 +8,7 @@ Text.defaultProps = {
   fontSize: 'text',
   textAlign: 'left',
   padding: 0,
-  margin: 'textMargin'
+  margin: 0
```
carlos-kelly: The idea here is to use an index value of the space scale instead of fixed values as defined by the default theme. That makes it nearly impossible for the user to easily override them.

src/theme/default-theme.js
```diff
@@ -23,10 +23,5 @@ export default {
     text: '44px',
     monospace: '20px'
   },
-  space: {
-    headerMargin: '16px 16px 24px',
-    textMargin: '16px',
-    listMargin: '16px 0',
-    slidePadding: '2em'
-  }
+  space: [16, 24, 32]
```
carlos-kelly: Instead of fixed values in a CSS string use consistent spacing as part of a scale that can be changed by the user.

docs/content/themes.md
```diff
@@ -0,0 +1,101 @@
...
+| `borderStyles`   | `border-style`                                                          |
+| `radii`          | `border-radius`                                                         |
+| `fonts`          | `font-family`                                                           |
+| `fontWeights`    | `font-weight`                                                           |
+| `lineHeights`    | `line-height`                                                           |
+| `letterSpacings` | `letter-spacing`                                                        |
+| `shadows`        | `box-shadow`, `text-shadow`                                             |
+| `zIndices`       | `z-index`                                                               |
+
+```
```
ryan-roemer: NIT: Extra fences here and on L101?

carlos-kelly: Good find!

PR #832, opened 2020-02-08
fix tutorial markdown link mismatch

johnstonmatt:
### Description

A relative path in a link to `tutorial.md` from `getting-started.md` was wrong, which resulted in a dead link to the tutorial.

#### Type of Change

- [x] This change is a documentation error fix

### How Has This Been Tested?

I clicked the link in preview mode, and it works now. :)


kale-stew:
Thank you for fixing this for us, @johnstonmatt!

Issue #834, opened 2020-02-11
Bug: Clicking the fullscreen icon bottom left of the screen in the JS example is not working. 


beautifwhale:
# Steps to reproduce

Create boilerplate project
```
spectacle-boilerplate \
  --mode "js" \
  --name "deck-o-awesomeness" \
  --description "My fantastic Spectacle presentation" \
--dir presentation
```

Install dependencies
```
cd presentation && npm i
```

Start presentation
```
npm start
```

# Expected

Presentation starts, clicking fullscreen sets the presentation to fullscreen. 

# Actual

Clicking the fullscreen icon bottom left of the screen in the JS boilerplate is not working.

ryan-roemer:
Thanks @beautifwhale !

This actually looks like an upstream issue with spectacle, so we've transferred it here. Issue is reproducible on the auto-presentation deploy here: https://spectacle.formidable.com/?slide=0&slideElement=-1

ryan-roemer:
Oh, and it's a bit confusing, but if you hit other issues with the CLI tool `spectacle-boilerplate`, it's actually part of the `spectacle-cli` package (that provides `spectacle` and `spectacle-boilerplate`). No worries here -- we've found the right home for the issue.

kale-stew:
Closed by #835

PR #835, opened 2020-02-11
Fixed Full Screen

carlos-kelly:
- Fixes zIndex issue which broke the full screen button on the example template
- Adds support for full screen in Safari

#834


ryan-roemer:
Works in Chrome. 

I'm getting `Browser does not support Presentation API` in FF `72.0.2 (64-bit)`. I'm updating FF to see if that helps.

ryan-roemer:
Same error for FF `73.0 (64-bit)`.

PR #850, opened 2020-02-28
Remove lineHeight unit from notes text

treyhoover:
### Description

Fixes an issue with notes overlapping in presenter view. (See image below for bug)

#### Type of Change

image

### DIFF
--- a/src/components/deck/presenter-deck.js
+++ b/src/components/deck/presenter-deck.js
@@ -164,7 +164,7 @@ const PresenterDeck = props => {
           <Timer />
         </Box>
         <NotesContainer>
-          <Text fontFamily={SYSTEM_FONT} lineHeight="180%" fontSize="1.5vw">
+          <Text fontFamily={SYSTEM_FONT} lineHeight="1.2" fontSize="1.5vw">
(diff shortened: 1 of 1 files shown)


treyhoover:
The issue is highlighted here: https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#Prefer_unitless_numbers_for_line-height_values

@ryan-roemer, This is only an issue when using `spectacle-mdx-loader` because the generated html is slightly different (each note has an extra div). This change should resolve that, but we may want to create a separate issue for that project to standardize the html outputted.

ryan-roemer:
@treyhoover -- Let's maybe also file the `spectacle-mdx-loader` issue with a lower priority just to track it as it would be nice to harmonize the outputs across JSX/MD/MDX modes.

treyhoover:
> @treyhoover -- Let's maybe also file the `spectacle-mdx-loader` issue with a lower priority just to track it as it would be nice to harmonize the outputs across JSX/MD/MDX modes.

Sure thing. Captured [here](https://github.com/FormidableLabs/spectacle-mdx-loader/issues/3).

Issue #857, opened 2020-03-06
Docs: Postcss in react-static@5 fails CSS building


ryan-roemer:
Probably will go away with #838 upgrade to react-static@7, but for how we're hitting this:

```sh
$ yarn prod:build
ERROR in ./node_modules/prismjs/themes/prism.css
Module build failed: ModuleBuildError: Module build failed: BrowserslistError: Unknown browser query `dead`
    at error (/Users/rye/scm/fmd/spectacle/docs/node_modules/browserslist/index.js:37:11)
    at /Users/rye/scm/fmd/spectacle/docs/node_modules/browserslist/index.js:222:9
    at Array.forEach (<anonymous>)
    at browserslist (/Users/rye/scm/fmd/spectacle/docs/node_modules/browserslist/index.js:196:13)
    at cleanBrowsersList (/Users/rye/scm/fmd/spectacle/docs/node_modules/caniuse-api/dist/utils.js:56:59)
    at setBrowserScope (/Users/rye/scm/fmd/spectacle/docs/node_modules/caniuse-api/dist/index.js:29:43)
    at Object.<anonymous> (/Users/rye/scm/fmd/spectacle/docs/node_modules/caniuse-api/dist/index.js:91:1)
    at Module._compile (internal/modules/cjs/loader.js:955:30)
    at Module._extensions..js (internal/modules/cjs/loader.js:991:10)
    at Object.require.extensions.<computed> [as .js] (/Users/rye/scm/fmd/spectacle/docs/node_modules/babel-register/lib/node.js:152:7)
    at /Users/rye/scm/fmd/spectacle/docs/node_modules/webpack/lib/NormalModule.js:195:19
    at /Users/rye/scm/fmd/spectacle/docs/node_modules/loader-runner/lib/LoaderRunner.js:367:11
    at /Users/rye/scm/fmd/spectacle/docs/node_modules/loader-runner/lib/LoaderRunner.js:233:18
    at runSyncOrAsync (/Users/rye/scm/fmd/spectacle/docs/node_modules/loader-runner/lib/LoaderRunner.js:143:3)
    at iterateNormalLoaders (/Users/rye/scm/fmd/spectacle/docs/node_modules/loader-runner/lib/LoaderRunner.js:232:2)
    at iterateNormalLoaders (/Users/rye/scm/fmd/spectacle/docs/node_modules/loader-runner/lib/LoaderRunner.js:221:10)
[...]

PR #858, opened 2020-03-06
BUG: Prism production build failure + misc.

ryan-roemer:
## Status
- [x] Undo Travis on `task/rewrite` branch (I've re-enabled this temporarily to get surge deploy).

## Work
- Bandaid prism postcss minification bug by disabling CSS minification. For #857 as a temporary measure.
- Add missing `docs/dist-prod/open-source/spectacle` symlink needed for `yarn prod:serve`
- Catch `yarn.lock` up. This should fail yarn install in CI, so not sure if this is something just on my machine as I didn't change deps...
- Lint was super slow, so I updated lints so docs one is more honed (had duplicates) and root project one now only lints specific dirs (it was doing way too many after docs addition of things that were not real source) and does the root + docs so you only really need the root one.


ryan-roemer:
Staging site looks good: http://formidable-com-spectacle-staging-858.surge.sh/open-source/spectacle/docs

Going to now disable Travis integration (which should leave the staging site).

PR #866, opened 2020-03-14
fix codepane ignores props.theme

### LINKED ISSUE #865: vsDark theme is hardcoded in CodePane source
CodePane component does not pass theme prop to the underlying Highlight component. So passing any theme prop does not have any effect on the CodePane.

Turanchoks:
Fixes #865


kale-stew:
Thank you, @Turanchoks!

PR #876, opened 2020-04-13
Added touch controls to navigate presentations

Yamanqui:
### Description

On mobile devices, without keyboard, it was impossible to move through the presentations. Users would expect being able to swipe left or right to advance and go back.

I copied the way `use-keyboard-controls.js` handled it, replacing the keyboard with touch events.

On browsers without touch capabilities, the events never fire, so having the hooks should not change anything.

#### Type of Change

### How Has This Been Tested?

I ran the development server and tested on both my desktop and mobile devices, using the example presentations.


kale-stew:
Awesome, thank you for this @Yamanqui!

PR #930, opened 2020-07-10
Adds business logic hooks and util functions

carlos-kelly:
To our main-7 branch this adds a bunch of the business logic hooks and util functions from monocle. The major difference is replacing lodash functions with native ones.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,14 +42,19 @@
   "browserslist": "> 0.25%, not dead",
   "dependencies": {
     "broadcastchannel-polyfill": "^1.0.0",
+    "deepmerge": "^4.2.2",
     "history": "^4.9.0",
     "marksy": "^8.0.0",
     "normalize-newline": "^3.0.0",
     "prism-react-renderer": "^0.1.7",
     "query-string": "^6.8.2",
+    "react-fast-compare": "^3.2.0",
     "react-spring": "^8.0.25",
     "styled-components": "^4.3.2",
     "styled-system": "^5.1.1",
+    "ulid": "^2.3.0",
+    "use-immer": "^0.4.0",
+    "use-resize-observer": "^6.1.0",
     "wonka": "^3.2.0"
   },
   "peerDependencies": {
--- a/package.json
+++ b/package.json
@@ -42,9 +42,9 @@
   "browserslist": "> 0.25%, not dead",
   "dependencies": {
     "broadcastchannel-polyfill": "^1.0.0",
-    "deepmerge": "^4.2.2",
     "history": "^4.9.0",
     "marksy": "^8.0.0",
+    "merge-anything": "^3.0.3",
     "normalize-newline": "^3.0.0",
     "prism-react-renderer": "^0.1.7",
     "query-string": "^6.8.2",
--- a/src/hooks/use-action-dispatcher.js
+++ b/src/hooks/use-action-dispatcher.js
@@ -1,0 +1,17 @@
+function identity(sender) {
+  return sender;
+}
+
+export default function useActionDispatcher(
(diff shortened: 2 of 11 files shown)


src/hooks/use-location-sync.js
```diff
@@ -0,0 +1,92 @@
+import { createBrowserHistory } from 'history';
+import QS from 'query-string';
+import isEqual from 'react-fast-compare';
+import merge from 'deepmerge';
+
+// Needed to properly merge query strings. (Hook consumers can also provide
+// their own merge function if necessary)
+function defaultMergeLocation(object, ...sources) {
+  return mergeWith(object, ...sources, (left, right, key) => {
```
ryan-roemer, line 9: I don't see `mergeWith` anywhere in scope here?

carlos-kelly: It's just a named curried function so 
`mergeLocation = defaultMergeLocation,` gets set in the function header

Then 
```
const nextLocation = mergeLocation(
        {},
        location,
        mapStateToLocation(initialState)
      );
      history.replace(nextLocation);
```
Tells me it just gets called as it is returned. I guess we don't need to name it.

src/hooks/use-deck-state.js
```diff
@@ -0,0 +1,88 @@
+import { useImmerReducer } from 'use-immer';
```
mhink, line 1: Note: I've been meaning to rewrite this to get rid of the `use-immer` dependency- happy to PR that sooner rather than later

src/hooks/use-steps.js
```diff
@@ -0,0 +1,165 @@
+import * as React from 'react';
+import { ulid } from 'ulid';
+import * as transitionConfigs from 'transitions';
+import { useTransition } from 'react-spring';
+import { SlideContext } from 'Slide';
+import sortBy from '../utils/sort-by';
```
mhink, line 6: double import here?

src/hooks/use-action-dispatcher.js
```diff
@@ -0,0 +1,17 @@
+function identity(sender) {
+  return sender;
+}
+
+export default function useActionDispatcher(
```
mhink, line 5: This function could probably be inlined into `use-deck-state` or removed altogether (in favor of directly using `React.useCallback`).

src/hooks/use-deck-state.js
carlos-kelly: If you want to PR that on monocle I can move that over. Immer is not a huge concern to me right now. So no rush.

src/hooks/use-steps.js
carlos-kelly: Good call!

src/hooks/use-location-sync.js
ryan-roemer: I don't think I explained that right... 😉   I mean the exact characters `mergeWith` do not occur anywhere else within this file at https://github.com/FormidableLabs/spectacle/blob/7918632c58ef3482776f257da70666f7716f0388/src/hooks/use-location-sync.js

carlos-kelly: Oooh you are right. We are missing that import from lodash. I will look into that before merging this.

package.json
```diff
@@ -44,12 +44,17 @@
     "broadcastchannel-polyfill": "^1.0.0",
     "history": "^4.9.0",
     "marksy": "^8.0.0",
+    "merge-anything": "^3.0.3",
```
ryan-roemer, line 47: Was taking a peek at this and they just updated to `3.0.4` so might just want to catch it, but no biggie.

Issue #928, opened 2020-07-09
Feature request: Copy-and-pasteable code blocks.


ryan-roemer:
### Description

Our code blocks like https://spectacle.formidable.com/?slide=3&slideElement=-1 visually "look" like normal code, but you can't copy and paste them straight up -- the line numbers are part of the paste and all indentation is removed.

### Proposal

Refactor code blocks to additionally be copy-and-pasteable in exactly the same format as originally provided, indents and all, without line numbers. (Basically, like GitHub code works).

ryan-roemer:
*Update*: current WIP PR fixes this with copy and pasteable code blocks that also skip the numbers in MD (I've highlighted some code and spacing in this screenshot) -- https://github.com/FormidableLabs/spectacle/pull/934

![Screen Shot 2020-07-28 at 11 05 58 AM](https://user-images.githubusercontent.com/445557/88704267-b0671580-d0c2-11ea-92c9-486ef8bac03c.png)

Issue #898, opened 2020-05-05
Code block indentation spacing is too small.


ryan-roemer:
See: https://github.com/FormidableLabs/spectacle/issues/886#issuecomment-619770673

**Actual Code**:
![Screenshot 2020-04-27 at 08 55 15](https://user-images.githubusercontent.com/8681413/80342686-dcc18a80-8864-11ea-8ae7-9143e80df58f.png)

**Presentation**:
<img width="329" alt="Screenshot 2020-04-27 at 08 55 22" src="https://user-images.githubusercontent.com/8681413/80342697-e2b76b80-8864-11ea-94b9-dcbfde3b5259.png">

I've seen this too in private presentations using MDX.

LongLiveCHIEF:
I'm getting this when using JSX, so I don't think it's just md or MDX.  even when using `indentSize` prop.  I also tried switching my `CodePane` content so it was using tabs instead of spaces, and it still trimmed all tab characters and whitespace upon presentation.

ryan-roemer:
@LongLiveCHIEF -- Is it trimming them, or are they just really, really small? I ended up doing like 8x the normal spaces I wanted then, finally saw them. (My strong hunch is that it's styling related)

LongLiveCHIEF:
I'll check back in the morning and let you know.

LongLiveCHIEF:
@ryan-roemer it's actually a combination of both.  There is no whitespace or html space characters in the markup, and the padding is 5px + 0.25em of the flex container.

LongLiveCHIEF:
Whitespace/tab characters shouldn't be stripped, because they need to be there for copy/paste capabilities for the presentation visitors to copy code into their IDE's without it looking like a mess.

ryan-roemer:
@LongLiveCHIEF -- My understanding is that the whitespace is "translated" by the code formatter to then a style amount, so it's not "stripped" per se, it's just "used for math" to do the style part.

Let me rephrase my question -- if you do 8x the normal whitespacing in code samples, do you visually see _more_ indentation or absolutely no difference in slide results?

ryan-roemer:
I've captured "copy and pasteable code blocks" separately as a feature request in: https://github.com/FormidableLabs/spectacle/issues/928

ryan-roemer:
*Update*: current WIP PR fixes this with correct spacing (and it's real spaces too) in MD -- https://github.com/FormidableLabs/spectacle/pull/934

[...]

PR #934, opened 2020-07-28
v7: Integrate Markdown, Markdown Slide Set, and new Code Pane

carlos-kelly:
- Integrates new Markdown support for v7
- Updates CodePane to support our theming constructs
- Updates MD and JS examples with a CodePane

**JS**

js-example

**Markdown**

markdown-example-2

markdown-example


examples/md/index.js
```diff
@@ -1,10 +1,11 @@
 import React from 'react';
 import { render } from 'react-dom';
 
-import { Box, Deck, FlexBox, FullScreen, Markdown, Progress } from 'spectacle';
+import { Box, Deck, FlexBox, FullScreen, Progress } from 'spectacle';
 
 // SPECTACLE_CLI_MD_START
 import mdContent from './slides.md';
+import { MarkdownSlideSet } from '../../src/components/markdown/markdown';
```
ryan-roemer, line 8: 1. This will need to come off a root level `spectacle` re-export.
2. It cannot be within the `SPECTACLE_CLI_MD_START` + `SPECTACLE_CLI_MD_END` control comments. These get regex-replaced by the CLI for custom slides.

examples/md/slides.md
```diff
@@ -44,7 +44,7 @@ Typography
 ```js
 class SuperCoolComponent extends React.Component {
   render() {
-    return <p>code slide works in markdown too whaaaaat</p>;
+    return <p>code slide works in markdown too whaaaaat</p><p>code slide works in markdown too whaaaaat</p><p>code slide works in markdown too whaaaaat</p>;
```
ryan-roemer, line 47: NICE -- this looks great for the open issues we have for:

* MD code indentation bug: #898 
* Copy and pasteable code blocks: #928

LongLiveCHIEF: say _whaaaaat_? 😛 Let me know if you want me to do any code reviews or toss it through my use case when it's ready to go.

src/components/markdown/markdown.js
```diff
@@ -0,0 +1,234 @@
...
+
+  return (
+    <TemplateComponent {...restProps}>
+      {children}
+      {noteElements}
+    </TemplateComponent>
+  );
+};
+
+export default Markdown;
```
ryan-roemer, line 144: Can we just _not_ have a default export since we have multiple exports already?

src/utils/remark-rehype-presenter-notes.js
```diff
@@ -0,0 +1,27 @@
+import zone from 'mdast-zone';
+import unistVisit from 'unist-util-visit';
+import * as mdast from 'mdast-builder';
+
+export default function remarkRehypePresenterNotes(noteCallback) {
+  return tree => {
```
ryan-roemer, line 6: Can we maybe move the functional code to the bottom of this function?

It was hard for me at least to understand it on first glance, to figure out exactly what those functions below are, that they're hoisted, and then that they're used above.

carlos-kelly: Sounds good!

carlos-kelly:
@ryan-roemer That should fix those two things! Thanks for the review.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "spectacle",
-  "version": "6.2.0",
+  "version": "7.0.0-alpha.1",
```
ryan-roemer, line 3: Shouldn't we do this on `npm version` command instead?

PR #941, opened 2020-08-15
Give slide backgroundColor to template

yag00:
### Description

Would it be possible to give the slide backgroundColor to the template. 
It is usefull to have your template dynamically adapt when you alternate between dark and light slide

#### Type of Change

- New feature (non-breaking change which adds functionality)
- Very minor change : It just pass an extra property (backgroundColor) to the template object.

### DIFF
--- a/src/components/slide.js
+++ b/src/components/slide.js
@@ -203,7 +203,8 @@ const Slide = props => {
         {typeof template === 'function' &&
           template({
             slideNumber: slideNum,
-            numberOfSlides: numberOfSlides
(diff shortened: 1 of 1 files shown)


fritz-c:
There are existing ways to accomplish what this feature would aim to accomplish ([example](https://codesandbox.io/s/track-background-color-spectacle-y4iq9p)), so I'm going to close this one.

fritz-c:
Thank you for the time you put into contributing to the library.

PR #942, opened 2020-08-16
Use Broadcast Channel for Dual-Browser message bus for presenting.

carlos-kelly:
Adds the Broadcast Channel hook to function as a message bus between the Audience and Presenter View

New Feature! The URL on the audience window also updates in real time!

Kapture 2020-08-15 at 18 58 38

### DIFF
--- a/src/components/deck/default-deck.js
+++ b/src/components/deck/default-deck.js
@@ -1,0 +1,28 @@
+import * as React from 'react';
+import Deck from './deck';
+import useBroadcastChannel from '../../hooks/use-broadcast-channel';
+
+export default function DefaultDeck(props) {
+  const deck = React.useRef();
+
+  const [postMessage] = useBroadcastChannel(
+    'spectacle_presenter_bus',
+    message => {
+      if (message.type !== 'SYNC') return;
+      const nextView = message.payload;
+      if (deck.current.initialized) {
+        deck.current.skipTo(nextView);
+      } else {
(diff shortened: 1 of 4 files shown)


src/components/deck/default-deck.js
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+import Deck from './deck';
+import useBroadcastChannel from '../../hooks/use-broadcast-channel';
+
+export default function DefaultDeck(props) {
+  const deck = React.useRef();
+
```
carlos-kelly, line 7: Default Deck is just a wrapper that adds the broadcast channel logic to the deck's navigation. We don't want it as part of the internal deck component since those are used on overview and presenter modes and do not need this logic.

src/hooks/use-broadcast-channel.js
```diff
@@ -0,0 +1,62 @@
...
+    userMessageHandlerRef.current = onMessage;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [...deps, postMessage]);
+
+  React.useEffect(() => {
+    if (!channel) return;
+    const messageHandler = event => {
+      const rawMessage = event.data;
+      const message = JSON.parse(rawMessage);
+      // console.log("RX:", message.type, message.payload);
```
ryan-roemer, line 52: MINOR: Remove debug comment?

PR #946, opened 2020-09-10
Feature/templates progress v7

carlos-kelly:
* Adds support for Templates
* Fixes Progress component
* Adds standard template to Example JS

image


ryan-roemer (approved):
LGTM!

Let's get CI green once the prettier vs eslint battle is wrangled.

PR #947, opened 2020-09-11
Update ESLint Configuration; Delete Older Files

carlos-kelly:
- Fixes ESLint configuration, fixes missed lint issues
- Deleted two older files
- Upgraded deps


docs/src/components/mdx.js
```diff
@@ -305,7 +305,7 @@ const AnchorIcon = styled(SvgAnchor)`
   height: 100%;
 `;
 
-const Header = tag => ({ id, children }) => {
+const Header = ({ tag, id, children }) => {
```
carlos-kelly, line 308: There is not a great way to assign propTypes to a curried component so I flattened it and in the component map, spread the tag and incoming props.

.eslintrc
```diff
@@ -35,9 +37,6 @@
     }
   },
   "rules": {
-    "prettier/prettier": [
-      "error"
-    ],
```
carlos-kelly, line 40: Recommended config does this for us, no need.

ryan-roemer (approved):
LGTM

PR #950, opened 2020-09-17
Updated Overview Mode

carlos-kelly:
- Added `Overview` mode for v7
- Added `BroadcastChannel` polyfill
- Ensured line endings are in `lf`
- Overview now has focus support for tabbing to a slide
- Selecting a slide will come in a future PR

Screen Shot 2020-09-17 at 9 41 45 AM
Screen Shot 2020-09-17 at 9 41 51 AM


examples/js/index.js
```diff
@@ -19,8 +19,13 @@ import {
   Box,
   Image,
   CodePane,
+  Markdown,
   Notes
 } from 'spectacle';
+import {
+  MarkdownSlide,
+  MarkdownSlideSet
+} from '../../src/components/markdown/markdown';
```
ryan-roemer, line 28: This isn't going to work for one-page.

carlos-kelly: Yup
Once again it's the auto import.

src/hooks/use-slides.js
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { ulid } from 'ulid';
 
-const PLACEHOLDER_CLASS_NAME = 'slide-placeholder';
+export const PLACEHOLDER_CLASS_NAME = 'spectacle-v7-slide';
```
ryan-roemer, line 4: Do we need a TODO/ticket to replace this before merge to main?

carlos-kelly: `PLACEHOLDER_CLASS_NAME` is probably not the right variable name. It's essentially the class selector we apply to every `<Slide />` `div` no matter the origin, JSX, MDX, MD, SlideSet. Maybe `SPECTACLE_SLIDE_SELECTOR`, I'm unsure if it should be a randomly generated ID (that would be consistent through the lifecycle of the deck) or just a string.

ryan-roemer: I'm ok with variable name, I meant `spectacle-v7-slide`. Are we intending to change this every semver major for some reason?

ryan-roemer: But go ahead with var name change if you think that's better!

ryan-roemer (approved):
Looks good! I had some extra comments in slack about the overall functioning of overview mode, but I don't think they block this PR

PR #955, opened 2020-10-17
Support presenter and overview mode query params inside the url.

carlos-kelly:
- Adds support for `overviewMode=true` or `presenterMode=true` to set the deck style into those modes.
- Supports ignoring the param if the value is `false`.


ryan-roemer (approved):
Code LGTM. Haven't tested out changes

upatel32 (approved):
💯

PR #962, opened 2020-11-24
v7: Bump version number for Babel to build on Windows

carlos-kelly:
Bumps Babel version numbers which allows for functional building on Windows CMD/PowerShell

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,12 +75,12 @@
     "react-dom": "^16.9.0"
   },
   "devDependencies": {
-    "@babel/cli": "^7.4.4",
-    "@babel/core": "^7.4.5",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "spectacle",
-  "version": "6.2.0",
+  "version": "7.0.0-beta.1",
```
ryan-roemer, line 3: Do we want to do this straight from `main-7` for `beta|next` npm publish? (Don't mind here or in main-7, just thought I'd ask).

ryan-roemer (approved):

package.json
carlos-kelly: Ah yes, good point!

PR #967, opened 2020-12-13
Export and Print Mode

carlos-kelly:
Adds export and print mode to Spectacle v7

Export Mode: A PDF-exportable version of slides in sequential order with the default theme
Print Mode: A PDF-exportable (and printable) version of slides in sequential order using the print theme

This also adds bi-directional syncing for query params for different modes and keyboard shortcuts.

Export Mode Left, Print Mode Right

Screen Shot 2020-12-13 at 2 30 12 PM

The goal is to ensure each slide is best scaled per-page by default. Inside Chrome you can then choose to put multiple pages on a slide. Previously slides could get cut off so this is a fix compared to v6.
Screen Shot 2020-12-13 at 2 34 10 PM

### DIFF
--- a/src/components/deck/index.js
+++ b/src/components/deck/index.js
@@ -57,6 +57,8 @@ export default function SpectacleDeck(props) {
         toggleMode(e, SPECTACLE_MODES.PRESENTER_MODE),
       [KEYBOARD_SHORTCUTS.PRINT_MODE]: e =>
         toggleMode(e, SPECTACLE_MODES.PRINT_MODE),
+      [KEYBOARD_SHORTCUTS.EXPORT_MODE]: e =>
+        toggleMode(e, SPECTACLE_MODES.EXPORT_MODE),
       [KEYBOARD_SHORTCUTS.OVERVIEW_MODE]: e =>
         toggleMode(e, SPECTACLE_MODES.OVERVIEW_MODE)
     },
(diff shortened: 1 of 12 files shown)


src/components/deck/index.js
```diff
@@ -1,36 +1,52 @@
...
         return;
       }
-      setMode(newMode);
+
+      mode.current = newMode;
+
+      location.search = stringifyQS({
+        slideIndex,
+        stepIndex,
+        ...modeSearchParamForKey(newMode)
```
carlos-kelly, line 48: For every mode transition we will update the `location` object so it will be also tracked in history. This also preserves existing params as well.

src/print-mode/index.js
```diff
@@ -0,0 +1,45 @@
...
+`;
+
+const PrintStyle = createGlobalStyle`
+  @media print {
+    body, html {
+      margin: 0;
+    }
+    ${AnimatedDiv} {
+      @page {
+        size: letter landscape;
```
carlos-kelly, line 18: In a follow up PR, these settings will be props so users can customize the print experience. Print queries do have to be global styles so our flexibility is limited.

src/utils/constants.js
```diff
@@ -4,8 +4,10 @@ export const SYSTEM_FONT =
   '-apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Helvetica Neue", Helvetica, sans-serif';
 
 export const KEYBOARD_SHORTCUTS = {
-  PRESENTER_MODE: 'mod+p',
-  OVERVIEW_MODE: 'mod+o',
+  PRESENTER_MODE: 'mod+shift+p',
+  OVERVIEW_MODE: 'mod+shift+o',
+  PRINT_MODE: 'mod+shift+r',
+  EXPORT_MODE: 'mod+shift+e',
```
carlos-kelly, line 10: I'm not sold on these but it's best to make most include <kbd>shift</kbd> since many browsers do things on just <kbd>meta</kbd>.

ryan-roemer (approved):
Looks great generally! Two questions:

- if you're printing you probably don't want gray backgrounds like your example. I could select to omit backgrounds in print dialog for no background, but should we just do this in printMode?
- What is the guidance for getting rid of the white bar at the bottom of printing to letter size? I wasn't clear if we need other export params or maybe custom print paper sizes?

carlos-kelly:
@ryan-roemer 

1. Good idea, I'll look into that. Might just be what our print theme is right now but the background should absolutely be white.
2. We best-fit scale, I'd love to center it on a page but that's proving to be inconsistent. Our default slide size is a `16:9` ratio while a letter page is `5:4`.

PR #977, opened 2021-01-28
Add Auto-Play support.

carlos-kelly:
- Adds auto play support with props for looping and setting a custom a custom duration.
- Adds documentation and unit tests.

### DIFF
--- a/src/utils/use-auto-play.js
+++ b/src/utils/use-auto-play.js
@@ -1,0 +1,28 @@
+import { useEffect, useRef } from 'react';
+import * as React from 'react';
+
+export const useAutoPlay = options => {
+  const {
+    enabled = false,
+    loop = false,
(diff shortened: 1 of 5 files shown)


src/utils/use-auto-play.js
```diff
@@ -0,0 +1,28 @@
...
+    } else {
+      navigation.stepForward();
+    }
+  });
+
+  useEffect(() => {
+    const tick = () => savedCallback.current();
+
+    if (enabled) {
+      const id = setInterval(tick, interval);
```
upatel32, line 24: small thing here, if `tick` is just calling `savedCallback.current`,  could we do `const id = setInterval(savedCallback.current, interval);`?

upatel32 (approved):

PR #978, opened 2021-01-29
Fix overflow in Presenter Mode for last slide.

carlos-kelly:
Currently Presenter Mode shows the last slide twice, this PR prevents the double rendering.

Screen Shot 2021-01-29 at 2 51 54 PM

### DIFF
--- a/src/components/presenter-mode/index.js
+++ b/src/components/presenter-mode/index.js
@@ -1,4 +1,5 @@
-import * as React from 'react';
+import React, { useRef, useCallback, useState, useEffect } from 'react';
+import styled from 'styled-components';
 import Deck from '../deck/deck';
 import { Text, SpectacleLogo } from '../../index';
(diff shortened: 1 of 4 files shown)


src/components/presenter-mode/index.js
```diff
@@ -108,15 +113,17 @@ export default function PresenterMode(props) {
...
-        <Deck
-          disableInteractivity
-          useAnimations={false}
-          backdropStyle={deckBackdropStyles.nextSlide}
-          ref={previewDeck}
-          theme={theme}
-        >
-          {children}
-        </Deck>
+        <PreviewSlideWrapper visible={showFinalSlide}>
```
carlos-kelly, line 116: The reason we use css `visibility` over not rendering it is we need to keep the `ref` to the preview deck always active.

upatel32 (approved):

Issue #985, opened 2021-02-23
Bug: setImmediate not defined in CodePane


pruesinghnagra:
Hi! I've been running into some issues trying to use `<CodePane>`. Is there an additional step that I need to do when installing to be able to use it?

I'm using React and npm. The error that I am getting is:

`Uncaught ReferenceError: setImmediate is not defined `

I'm also getting errors in the bundle.js compile about StyledComponents. I'm not using them but I can see that Spectacle is. I've had a look through the docs and can't seem to figure out if I should be installing more dependencies. 

Everything else seems to work great though! Just the` <CodePane>` component. Any help will be appreciated! Thanks :)

ryan-roemer:
Hi @pruesinghnagra ! Would it be possible for you to create a minimal public repository that exhibits the issue with build steps? In particular we want to look at your webpack version as webpack5 changed polyfills available by default. Thanks!

pruesinghnagra:
Ok sweet!!! Thanks @ryan-roemer :) Yeah I did run into some webpack problems actually! 

Here is a simple repo https://github.com/pruesinghnagra/lecture-template - hopefully it helps.

ryan-roemer:
Thanks @pruesinghnagra ! Confirmed. The bug is that we use a Node.js-only `setImmediate` that isn't in browsers and webpack previously polyfilled it but doesn't now in v5.

## Work
- [ ] Grep and remove all `setImmediate`/`clearImmediate` usage throughout.

## Temporary fix
In the meantime, you can polyfill `setImmediate` in your project with something like the following:

```diff
diff --git a/client/index.js b/client/index.js
index bea0314..214c159 100644
--- a/client/index.js
+++ b/client/index.js
@@ -1,3 +1,6 @@
+// Temporary fix (polyfill) for: https://github.com/FormidableLabs/spectacle/issues/985
+import "setimmediate";
+
 import React from 'react'
 import ReactDOM from 'react-dom'
 
diff --git a/package.json b/package.json
index 5dd3bca..b1cb7b0 100644
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "react-dom": "^17.0.1",
     "sass": "^1.29.0",
     "sass-loader": "^10.1.0",
+    "setimmediate": "^1.0.5",
     "spectacle": "^7.0.1",
     "spectacle-cli": "^0.6.3",
     "style-loader": "^2.0.0",
```

ryan-roemer:
[...]

PR #986, opened 2021-02-25
Swap setImmediate with requestAnimationFrame

carlos-kelly:
Swaps out `setImmediate` with `requestAnimationFrame`. Verified the scrolling for animation highlight groups still work and it scrolls to the correct spot for long CodePanes.

Addresses #985


ryan-roemer (approved):
Code LGTM! Haven't tested locally.

PR #990, opened 2021-03-10
fix: Ensure correct page size for export mode is default

carlos-kelly:
For `exportMode` which is designed for generating a PDF output of the deck, set the page size using CSS print media queries to the correct paper size which maps to the deck theme's slide size.

Screen Shot 2021-03-10 at 11 55 22 AM

### DIFF
--- a/src/components/deck/deck.js
+++ b/src/components/deck/deck.js
@@ -23,9 +23,15 @@ import {
   printWrapperStyle
 } from './deck-styles';
 import { useAutoPlay } from '../../utils/use-auto-play';
+import defaultTheme from '../../theme/default-theme';
 
 export const DeckContext = createContext();
 const noop = () => {};
+
+/**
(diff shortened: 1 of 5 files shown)


src/components/deck/deck.js
```diff
@@ -26,7 +26,7 @@ import { useAutoPlay } from '../../utils/use-auto-play';
 
 export const DeckContext = createContext();
 const noop = () => {};
-const DEFAULT_PRINT_SCALE = 0.773;
+const DEFAULT_PRINT_SCALE = 0.959;
```
carlos-kelly, line 29: The PDF DPI is `96`. We want to scale the slide down because it's a 1:1 px to 1/100th of an inch. However there are some unchangeable margins that make `0.96` too big. So we use `0.959` to prevent overflow.

ryan-roemer: Can we add that above in a code comment?

ryan-roemer (approved):
Looks great in my current deck!!!

PR #993, opened 2021-03-11
fixed link to contributing doc

jhumbug:
Fixed link to the contributing doc in the Tutorial


carlos-kelly (approved):

jhumbug:
After submitting this PR, I noticed that on staging the built docs are stripping .md from any markdown external link.  So not only is that contrib link still not exactly working because of that, any other links to markdown files in the docs are 404ing.

It seems to be because of a react-static plugin as described here: https://github.com/kitten/react-static-plugin-md-pages#relative-links-fixes.  I believe it's only supposed to strip relative links, but these aren’t relative and still getting stripped.  So we may want to address that before merging this in.

gksander (approved):
Nice, good catch!

gksander:
Uh oh, looks like Docusaurus found some broken links. If you run a local docusaurus build, you should be able to catch the broken links locally before it gets to CI.

fritz-c:
Yeah, I had been running it locally with the `docusaurus start` command but it didn't catch that double-`.md` one for whatever reason.

Issue #968, opened 2020-12-20
How to let list items "appear" with *.mdx


badabam:
### Question

I am running my slides from MDX and have trouble figuring out how to let items appear one by one.

#### Example
```
- this should appear first
- then this on the next ArrowRight press
- then this ...

---

- then the next slide should show

```

#### Attempts
I tried it on the stable release, but while trying to figure things out, I also looked into the current beta (Branch `main-7`) and  there is something that looks [like what I want.](https://github.com/FormidableLabs/spectacle/blob/cdd64740123b707e88ec79479ea4f9970792a9b5/src/components/markdown/markdown.js#L69) Unfortunately I can't get this working locally.

#### Tried custom solution
Not finding a solution in your source code (while still thinking it might exist), I created a Component to work with [(Link to code)](https://github.com/badabam/slides2021/blob/main/src/components/StepList.js), but it only works on the current stable release and breaks in `main-7`. It looks like you changed the way you render the slides (i.e. pages). 

### Help needed

I would love to hear from you to get this figured out – in any way you prefer. Could be a thread here, mail, zoom/google/teams-call.  If it helps, here is [the Repo I am working on](https://github.com/badabam/slides2021).

PR #994, opened 2021-03-12
Md list appear

gksander:
### Description

This PR adds a boolean prop `animateListItems` to the `Markdown` suite of components that automatically transforms MD list items into `Appear > ListItem`. This allows a user to provide MD like:

```jsx
<MarkdownSlide animateListItems>
{`
# Hey world

- One
- Two
- Three
`}
</MarkdownSlide>
```

and have the list items appear in one at a time.

This addresses issue #968.

#### Type of Change

### How Has This Been Tested?

I have added some Jest/Enzyme tests to test this new prop. I also added an example to the JS example deck that uses this new prop, which can be used to verify that the list items are animating in accordingly.

### DIFF
--- a/src/components/markdown/markdown.test.js
+++ b/src/components/markdown/markdown.test.js
@@ -1,0 +1,83 @@
+import React from 'react';
+import Enzyme, { mount } from 'enzyme';
+import { MarkdownSlide, MarkdownSlideSet } from './markdown';
+import Adapter from 'enzyme-adapter-react-16';
+import Deck from '../deck/deck';
+import { ListItem } from '../typography';
+import Appear from '../appear';
+
+Enzyme.configure({ adapter: new Adapter() });
+
+const mountInsideDeck = tree => {
+  return mount(<Deck>{tree}</Deck>);
+};
+
+describe('<MarkdownSlide />', () => {
+  it('should generate standard unordered lists by default', () => {
+    const wrapper = mountInsideDeck(
+      <MarkdownSlide>{`
+        - One
+        - Two
+        - Three
+      `}</MarkdownSlide>
+    );
+
+    expect(wrapper.find('ul')).toHaveLength(1);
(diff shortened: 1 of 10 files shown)


src/components/markdown/markdown.test.js
```diff
@@ -0,0 +1,83 @@
...
+    const wrapper = mountInsideDeck(
+      <MarkdownSlide animateListItems>{`
+        - One
+        - Two
+        - Three
+      `}</MarkdownSlide>
+    );
+
+    expect(wrapper.find('ul')).toHaveLength(1);
+    expect(wrapper.find(Appear)).toHaveLength(3);
```
gksander, line 40: I'd love to also test for the existence of `ListItem` components, but those only appear as you step through the slide. I wasn't sure if figuring out how to imperatively do that was worth the time. The enzyme tests are already in the realm of "testing implementation details", so this set of assertions felt sufficient.

In the future, we might consider moving to `react-testing-library` for testing to move away from testing implementation details. My ideal test would be along the lines of:

- Mount the slide.
- Check that none of the list items are present.
- Simulate a click.
- Check that the first list item is present (searching by text), but not the last two.
- Simulate two more clicks.
- Check that all list items are present (searching by text).

src/utils/mdx-component-mapper.js
```diff
@@ -1,19 +1,17 @@
 /* eslint-disable react/display-name */
 import * as React from 'react';
 import {
-  CodePane,
+  Heading,
```
gksander, line 4: There were some unused imports in this file. I usually clean this up with an editor action for "optimizing imports" (which clears those out). Looks like it also sorts them by alphabetical order, which is why this diff looks bigger than it should. (In effect, all that happened was `CodePan` and `CodeSpan` were removed.)

carlos-kelly (approved):
This looks fantastic!

PR #995, opened 2021-03-15
Migrate to Github Actions

jhumbug:
- Removes Travis
- Adds Github Actions
  - ci.yml - tests and builds core code on pushes and pull requests with changes to the core code
  - docs.yml - tests, builds and deploys docs on pushes and pull requests with changes to docs


package.json
```diff
@@ -12,7 +12,7 @@
     "check": "yarn lint && yarn prettier-check && yarn test",
     "check-ci": "yarn run check",
     "format": "yarn run lint-fix && yarn run prettier-fix",
-    "lint": "eslint '*.js' 'examples/**/*.js' src scripts 'docs/*.js' docs/src",
+    "lint": "eslint '*.js' 'examples/**/*.js' src scripts",
```
ryan-roemer, line 15: @jhumbug @carlos-kelly -- So the awkward thing we have right now is the play between root and docs eslints. The doc eslint only covers `*.js` which means that now we're missing coverage for `docs/src/`. Some thoughts:

- Do we want to have two lints? Or just pull out of docs and do all in root
- If we do have two lints, we need to make sure `docs/src` is covered as it isn't right now in this PR.

ryan-roemer (changes requested):
First pass done. Great work!

We have a "two lints" issue we should resolve. I'm leading towards "only lint in root, not docs", but open either way.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,52 @@
...
+      - '**'
+      - '!docs/**'
+
+jobs:
+  build:
+    name: Check and build codebase
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x, 14.x]
```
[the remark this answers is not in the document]
jhumbug: Added 15.x

.github/workflows/docs.yml
```diff
@@ -0,0 +1,72 @@
...
+jobs:
+  docs:
+    name: Build the docs
+    # TODO: Switch to `ubuntu-latest` when GH internal migration complete
+    # because it has awscli@2 built in.
+    # See `ubuntu-latest` note in: https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
```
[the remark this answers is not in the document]
jhumbug: added matrix with 15.x

package.json
jhumbug: As an FYI - The reason I made these changes stems from the separate workflows approach.  If you try to lint on the docs workflow, you only have the docs node_modules to work with. But the linter reaches up a directory and uses the eslintrc from the root.  That root eslintrc specifies the eslint-typescript parser and some eslint plugins that aren't in the docs' package json, so the lint doesn't work for docs since the root node_modules aren't installed.

I believe if we want to keep separate workflows and pull the lint out of docs and do it all in root, we'll have to install the root node_modules as part of the docs workflow in order to lint them when there are docs changes.  Should be doable.

I went ahead added 'src' to the docs linter so that should be covered if we decide to have two linters.

ryan-roemer: Oh, that makes sense! You know, one thing that might be good for harmonizing is doing and extends in the configuration of the root eslint  configuration instead of a completely new one

jhumbug: I went ahead and made the change so that there's only one linter again, while keeping separate gh workflows.  All I had to do was add the install of the root node_modules in the docs workflow.  Adds a few seconds to the ci/cd, but reduces complexity, so I think this is a good trade off.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Get Yarn Cache Directory Path
+        id: cache-dir-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+
+      - name: Use Yarn Cache
+        uses: actions/cache@v2
+        id: cache
+        with:
+          path: ${{ steps.cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ matrix.node-version }}-${{ hashFiles('**/yarn.lock') }}
```
ryan-roemer, line 40: Minor - will `**/yarn.lock` also include and hash `docs/yarn.lock`? Because we wouldn't want changes in `docs/yarn.lock` to invalidate the cache for this workflow which only deals with the root install.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,82 @@
...
+jobs:
+  docs:
+    name: Build the docs
+    # TODO: Switch to `ubuntu-latest` when GH internal migration complete
+    # because it has awscli@2 built in.
+    # See `ubuntu-latest` note in: https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        node-version: [15.x]
```
ryan-roemer, line 29: Let's pick latest stable (so `14.x`)

ryan-roemer (approved):
Great work @jhumbug ! Two minor things left to resolve, and one question open to @carlos-kelly

.github/workflows/ci.yml
jhumbug: Good eye.  I updated it to be specific.

.github/workflows/docs.yml
jhumbug: done!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,52 @@
+name: CI
+
+# Runs build and test on:
+#   every push that has a change in a file not in the docs folder
+#   every pull request with main branch as the base that has a change in a file not in the docs folder
+on:
+  push:
```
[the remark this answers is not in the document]
carlos-kelly: I think limiting to targeting main is fine.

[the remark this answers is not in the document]
jhumbug: Now limited to main

PR #996, opened 2021-03-17
MD component prop pass-thru

gksander:
### Description

This PR allows the user to pass a `componentProps` prop to the `Markdown` suite of components, and these props will be passed down to each individual component generated by the MD conversion process. E.g.,

```jsx
<MarkdownSlide componentProps={{ backgroundColor: 'red', color: 'blue' }}># Ugly header</MarkdownSlide>
```

will style the header with a color of blue, background of red. This gives the user some control over how their markdown components render elements.

#### Type of Change

### How Has This Been Tested?

Tests added to `markdown.test.js` to test these changes. You can also try the `componentProps` prop out within the JS example slides.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -105,19 +105,24 @@ declare module 'spectacle' {
     size: number;
   }>;
 
+  type MdComponentProps = { [key: string]: any };
+
   export const Markdown: React.FC<{
     animateListItems?: boolean;
     children: React.ReactNode;
+    componentProps?: MdComponentProps;
   }>;
 
   export const MarkdownSlide: React.FC<{
     animateListItems?: boolean;
     children: React.ReactNode;
+    componentProps?: MdComponentProps;
   }>;
 
   export const MarkdownSlideSet: React.FC<{
     animateListItems?: boolean;
     children: React.ReactNode;
+    componentProps?: MdComponentProps;
   }>;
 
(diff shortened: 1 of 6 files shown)


index.d.ts
```diff
@@ -105,19 +105,24 @@ declare module 'spectacle' {
     size: number;
   }>;
 
+  type MdComponentProps = { [key: string]: any };
```
gksander, line 108: I'm not thrilled about how loose this type is, but technically any props from valid MD components can be passed here, and the user can supply their own component map (in which case, we can't account for the prop types their components will accept). Seems like it'd be hard to make this as strict as possible - and making this unnecessarily strict is also a bad idea (since TS will yell about it not recognizing props that might actually be valid).

carlos-kelly: I also think leaving it open gives us API flexibility. We could also use the `Record` type but I think that's two sides of the same coin here.

carlos-kelly (approved):
This is great, but let's actually add back the example of this to `examples/js/index.js` perhaps on line 195.

ryan-roemer:
Great work @gksander and great suggestion @carlos-kelly ! 

One extra tip -- when changing `examples/js` make sure to run `yarn build-one-page` after and commit the changes to `example/one-page.html` which is generated from `examples/js`

carlos-kelly (approved):

PR #1006, opened 2021-04-07
CodePane Syntax Theme Import Fix

carlos-kelly:
### Description

This simplifies our theme imports to a single default dark theme and requires the user to import any custom themes in their own deck. This reduces the overall bundle size, but could potentially result in any usages of a string-based Prism theme `CodePane` to stop rendering.

Fixes # (issue)

#### Type of Change

### How Has This Been Tested?

To ensure the import doesn't break consumers that don't transform node modules, this was linked to a project to verify the cjs usage did not throw an `export` keyword error.


ryan-roemer (approved):
Code looks good. Let me know if you want me to try it out locally. 🚀

gksander (commented):
Code/docs look good to me! The only question I have is the type in `index.d.ts` for `CodePane`. The current type in there is:

```ts
Record<string, unknown> | string
```

Seems like we need to remove `string` as an option?

carlos-kelly:
Grant, good call! I will update the types. @ryan-roemer If you want to kick the tires that would be great! On my end this results in a reduction in bundle size.

PR #1011, opened 2021-04-21
add comments to better document hooks

soft-circles:
### Description

add comments to better document hooks

### DIFF
--- a/src/hooks/use-steps.js
+++ b/src/hooks/use-steps.js
@@ -11,7 +11,7 @@ const PLACEHOLDER_CLASS_NAME = 'step-placeholder';
  * Hook to determine whether or not a step is active or not. Note that
  * useCollectSteps() must be called prior to calling this method as useSteps()
  * requires activationThresholds to be available.
-*/
+ */
 export function useSteps(numSteps = 1, { id: userProvidedId, stepIndex } = {}) {
   const [stepId] = React.useState(userProvidedId || ulid);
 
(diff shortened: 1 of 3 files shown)


src/hooks/use-steps.js
```diff
@@ -7,7 +7,11 @@ import clamp from '../utils/clamp';
 
 const PLACEHOLDER_CLASS_NAME = 'step-placeholder';
 
-// TODO: Document this function really well- it's a public-facing API
+/*
+ * Hook to determine whether or not a step is active or not. Note that
```
mhink, line 11: This isn't quite accurate.

This hook is used to create components which can 'participate' in a presentation (the canonical example is `<Appear />`, which uses this hook with `numSteps == 1`)

When a component uses this hook, it passes `numSteps`, which "reserves" that many steps within the slide progression. The logic in `<Deck>` and `<Slide>` takes this into account as the presenter steps forward and back through the presentation.

As for return values: the consuming component _must_ render the `placeholder` React element, because that's how the "reservation" mechanism works.

The `step` return value indicates which _relative_ step the presentation has reached. That is, if a component requests 3 steps in the presentation, step will only ever have the values `-1` (if the presentation hasn't reached the stepper yet), `0`, `1`, or `2`.

`isActive` is really just an alias for `step > -1`.

---

Also, I think a better way of describing the context requirements here is to say that `activationThresholds` is _provided by_ `<Slide>` and _calculated_ by `useCollectSteps`. But I wouldn't mention it in user-facing documentation, since it's an implementation detail rather than an API.

soft-circles: Stealing this haha

carlos-kelly (approved):

PR #1016, opened 2021-04-29
export useMousetrap hook for consumers of the library

soft-circles:


carlos-kelly:
Let's also add typescript types otherwise we won't be able to import this in a TypeScript project

soft-circles:
Oh forgot! Yes totally.

index.d.ts
```diff
@@ -11,6 +11,13 @@ declare module 'spectacle' {
     currentSlide: number;
   }) => React.ReactNode;
 
+  export type useMousetrap = (
+    keybinds: {
+      string: void;
+    }[],
+    deps: any[]
```
carlos-kelly, line 18: This is close!

keybinds should be
```
Record<string, (e: ExtendedKeyboardEvent, combo: string) => void>
```

You might need to import `ExtendedKeyboardEvent` from `mousetrap`.

carlos-kelly (approved):

Issue #1009, opened 2021-04-19
No Stepper Component? (is it now the use-steps hook?)


sneakers-the-rat:
### Prerequisites

**Feel free to delete this section if you have checked off all of the following.**

- [x] I've searched open [issues](https://www.github.com/FormidableLabs/spectacle/issues) to make sure I'm not opening a duplicate issue
- [x] I have read through the [docs](https://www.formidable.com/open-source/spectacle/docs) before asking a question
- [`^8.0.1`] I am using the latest version of Spectacle

### Describe Your Environment

What version of Spectacle are you using? (can be found by running `npm list spectacle`): 
`8.0.1`

What version of React are you using? (can be found by running `npm list react`): 
`17.0.2`

What browser are you using?
Firefox

What machine are you on?
macOS

### Describe the Problem

Stepper is [advertised on the docs](https://formidable.com/open-source/spectacle/docs/api-reference/#stepper), and is mentioned in the changelog, which links to PR https://github.com/FormidableLabs/spectacle/pull/843 which seems to add a `components/stepper.js` file, but that never made it to the `main` branch. Did this functionality get merged into some other object? In any case there's a mismatch between the docs and the code.

**Expected behavior:** [What you expect to happen]

`import { Stepper } from 'spectacle';`

then i guess being able to use it...

**Actual behavior:** [What actually happens]

`export 'Stepper' (imported as 'Stepper') was not found in 'spectacle' (possible exports: Appear, Box, CodePane, CodeSpan, Deck, DeckContext, FlexBox, FullScreen, FullSizeImage, Grid, Heading, Image, Link, ListItem, Markdown, MarkdownPreHelper, MarkdownSlide, MarkdownSlideSet, Notes, OrderedList, Progress, Quote, Slide, SlideContext, SpectacleLogo, Table, TableBody, TableCell, TableHeader, TableRow, Text, UnorderedList, defaultTheme, indentNormalizer, isolateNotes, mdxComponentMap, removeNotes)`

sneakers-the-rat:
Or, i guess more generally, what i'm trying to do is step through an animation of an SVG, and i want to hook into the keypress event so that the slide doesn't change but the animation proceeds.

sneakers-the-rat:
[...]

PR #1019, opened 2021-05-04
export hooks

sneakers-the-rat:
### Description

Export hooks! hooks are currently not exported from the main module. Attempting to import from within library fails, eg:

`import { useSteps } from 'spectacle/lib/hooks/use-steps';`

or making an alias in webpack like

```javascript
resolve: {
    alias: {
      '@useSteps': path.resolve(__dirname, '../node_modules/spectacle/lib/hooks/use-steps.js')
    }
```

causes errors like:

```
Uncaught TypeError: _React$useContext is null
    useSteps use-steps.js:69
```

Exporting them seems to solve this. This is the only way that i have found (looking through the source) to add custom step logic to a slide (eg. stepping through a series of callbacks to animate an svg), and the comments around the hooks seem to indicate they are intended to be public, but if there's some other way i should do this lmk

Fixes # (issue):
not necessarily a fix, but related to https://github.com/FormidableLabs/spectacle/issues/1009

#### Type of Change

- [x] This change requires a documentation update - same as https://github.com/FormidableLabs/spectacle/issues/1009

### How Has This Been Tested?

Doing something like this seems to create multiple steps, though i still haven't figured out exactly how to use them, following examples from `Appear` and `CodePane` 

```react
import { useSteps } from 'spectacle';

export default function SvgAnimator({
steps=[]
}){
  const numberOfSteps = steps.length;
  const { stepId, isActive, stepNum, placeholder } = useSteps(numberOfSteps, {stepIndex});
}
```

- [ ] I have commented my code, particularly in hard-to-understand areas - *not really necessary? pretty clear whats going on?*
- [ ] I have made corresponding changes to the documentation - *still unclear on how exactly hooks are intended to be used*
- [x] I have updated type definitions in `index.d.ts` for any breaking API changes - *N/A*
- [ ] I have added tests that prove my fix is effective or that my feature works - *still unsure how hooks are supposed to work*
- [ ] Any dependent changes have been merged and published in downstream modules - *not sure what this means!*

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -38,6 +38,19 @@ import { DeckContext } from './components/deck/deck';
 import useMousetrap from './hooks/use-mousetrap';
 import defaultTheme from './theme/default-theme';
 
+import useActionDispatcher from './hooks/use-action-dispatcher';
+import useAspectRatioFitting from './hooks/use-aspect-ratio-fitting';
+import useAutofillHeight from './hooks/use-autofill-height';
+import useBroadcastChannel from './hooks/use-broadcast-channel';
+import useDeckReducer from './hooks/use-deck-state';
+import { useToggleFullScreen } from './hooks/use-full-screen';
+import useKeyboardControls from './hooks/use-keyboard-controls';
+import useLocationSync from './hooks/use-location-sync';
+import usePresentation from './hooks/use-presentation';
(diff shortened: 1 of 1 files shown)


fritz-c:
Merged in with #1225 
Thank you for your contribution.

PR #1020, opened 2021-05-05
Add support for custom transitions.

carlos-kelly:
### Description

Adds support for custom transitions and two built-in transition objects: Fade and Slide. Slide is the default transitions for decks. Slides can have custom transitions that override the default Deck transition style on a per-slide basis.

- Updates one page and JS examples
- Updates documentation 

Fixes # (issue)

#### Type of Change

### How Has This Been Tested?

Added examples to both the OnePage and JS Decks.

### DIFF
--- a/src/components/slide/slide.js
+++ b/src/components/slide/slide.js
@@ -16,7 +16,6 @@ import { useSlide } from '../../hooks/use-slides';
 import { useCollectSteps } from '../../hooks/use-steps';
 import { GOTO_FINAL_STEP } from '../../hooks/use-deck-state';
 import { useSwipeable } from 'react-swipeable';
-import { defaultTransition } from '../transitions';
 
 const noop = () => {};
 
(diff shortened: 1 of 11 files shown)


src/components/slide/slide.js
```diff
@@ -92,6 +89,7 @@ export default function Slide({
   padding,
   textColor,
   template,
+  transition: slideTransition = {},
```
gksander, line 92: Should this be `defaultTransition` instead of `{}`? I see the import for `defaultTransition`, but don't see it being used - but I'd assume that'd go here?

gksander (approved):
This looks great! One small question.

src/components/slide/slide.js
carlos-kelly: Oh good catch, Slide already gets `defaultTransition` from deck. I can remove that import. But if we made it `defaultTransition` it would just assign the same thing to itself.

ryan-roemer (approved):
Code looks good. Haven't tried it out. As long as someone has verified one-page.html is still good I'm all good. Thanks!

carlos-kelly:
@ryan-roemer I did verify it because I gathered you'd probably want to leverage more discrete transitions in your deck going forward.

PR #1061, opened 2021-08-25
Fix Stepper component to apply (in)activeStyle properties properly

neobrain:
### Description

Stepper is a wrapper around SteppedComponent but failed to forward these arguments. This was not an issue for the Appear component since that one doesn't explicitly name its properties but just passes all of them to SteppedComponent through `...restProps`.

#### Type of Change

### How Has This Been Tested?

While working on a presentation of mine, I noticed the `activeStyle` property does not get taken into account. With this change, it does.

### DIFF
--- a/src/components/appear.js
+++ b/src/components/appear.js
@@ -104,6 +104,8 @@ export function Stepper({
       {...restProps}
       numSteps={values.length}
       alwaysAppearActive={alwaysVisible}
+      activeStyle={activeStyle}
+      inactiveStyle={inactiveStyle}
     >
       {(step, isActive) =>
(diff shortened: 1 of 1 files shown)


fritz-c:
Thank you for the PR, and apologies for taking this long to get around to checking it. It appears this issue may have been fixed in a random [eslint cleanup commit](https://github.com/FormidableLabs/spectacle/commit/605436120df0d5694d9a1210b9ea0d0675d5bc33#diff-98e68035b1055875ad6074bc26ef8661fd6cca37c8bd83ef5cc738b05976af9b) and is [working in v9.5.1](https://codesandbox.io/s/stepper-demo-spectacle-5epvw8)

PR #1072, opened 2021-11-04
Bug: Update docs link relativity

gingerrific:
### Description

Updates "Get Started" documentation link to be relative to page instead of relative to root.

#### Fixes
Current documentation link goes to `https://formidable.com/docs/basic-concepts/` which can fail if don't visit the base page first and even then seems to intermittently fail and show a 404 page. The link as is would also break if a user were to bookmark it. This change updates it to be `https://formidable.com/open-source/spectacle/docs/basic-concepts/`

#### Type of Change

### How Has This Been Tested?

Locally confirmed link works as expected


carlos-kelly (approved):
Works great!

PR #1074, opened 2021-11-09
Chore: prune obsolete hooks

scottrippey:
### Description

Removes unused files.

#### Type of Change

- [x] Refactoring (non-breaking code improvement, cleanup, organization, or optimization)

### How Has This Been Tested?

Example slides have been regression tested.  Touch and keyboard navigation has been validated.

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have updated type definitions in `index.d.ts` for any breaking API changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Any dependent changes have been merged and published in downstream modules


scottrippey:
Noteworthy changes; I validated that references are gone to all these files.
Additionally:
`use-keyboard-controls` is no longer used; we use `mousetrap` now
`use-touch-controls` is no longer used; we use `react-swipeable` now

ryan-roemer:
I believe that there are now unused production dependencies from the three factor. Can we scour generally for unused dependencies?

package.json
```diff
@@ -68,8 +66,7 @@
     "ulid": "^2.3.0",
     "unified": "^9.0.0",
     "unist-util-visit": "^2.0.3",
-    "use-resize-observer": "^6.1.0",
-    "wonka": "^3.2.0"
```
scottrippey, line 72: Wonka was being used in `use-transition-pipe.js`, which was deleted.

package.json
```diff
@@ -45,13 +45,11 @@
     "broadcastchannel-polyfill": "^1.0.0",
     "dedent": "^0.7.0",
     "history": "^4.9.0",
-    "marksy": "^8.0.0",
     "mdast-builder": "^1.1.1",
     "mdast-util-assert": "^3.0.0",
     "mdast-zone": "^4.0.0",
     "merge-anything": "^3.0.3",
     "mousetrap": "^1.6.5",
-    "normalize-newline": "^3.0.0",
```
scottrippey, line 54: Did not find any usages of this lib in the codebase.  Does it seem safe to remove?  (Removing dependencies is hard without full context!)

package.json
```diff
@@ -45,13 +45,11 @@
     "broadcastchannel-polyfill": "^1.0.0",
     "dedent": "^0.7.0",
     "history": "^4.9.0",
-    "marksy": "^8.0.0",
```
scottrippey, line 48: Did not find any usages of this lib in the codebase.  Does it seem safe to remove?  (Removing dependencies is hard without full context!)

scottrippey: After investigating the git history, looks like this lib was used for an old Markdown component.  
So yes, safe to delete now!

package.json
scottrippey: Looks like this lib was used for an old Deck component.
So safe to delete too!

PR #1075, opened 2021-11-17
v9: Typescript Conversion

scottrippey:
### Description

Converts codebase to TS.
Attempts to minimize functional changes

Only converted the `/src` folder; did not convert the `/docs` folder.

#### Type of Change

- [x] Chore (refactor, optimization)

### How Has This Been Tested?

Test the app for any regressions!  No changes should be visible.

Steps to test:
1. Checkout branch, `yarn install` of course
2. Run `yarn checks` and verify all pass
3. Run `yarn start:js` and verify that the sample deck works without any differences
4. Run `yarn start:md` and verify nothing has changed

Additional testing with `spectacle-visual-editor`:
1. Checkout `spectacle-visual-editor` as a sibling to `spectacle`
2. In `spectacle`, run `yarn link`  and `yarn build`
3. Clear out `node_modules` from spectacle (this might be required to avoid conflicts with `styled-components`)
4. Switch to `spectacle-visual-editor` and run `yarn link spectacle` then `yarn install`
5. Run `yarn start` to bring up the visual editor
6. Ensure the editor works as expected
7. Run `yarn typecheck` and verify types are working


src/components/image.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const Image = styled.img(compose(layout, position)) as ImageType;
+const FullSizeImage = (styled(Image) as unknown) as ImageType;
+
+FullSizeImage.defaultProps = {
+  maxWidth: '100%',
+  maxHeight: '100%'
+};
+
+export { Image, FullSizeImage };
```
scottrippey, line 16: Note:
When renaming files, if 50% of the file changes, then Git marks it as a "delete + create" instead of a rename.
Since this file was so small, it looks like the (minor) TS conversion went past that 50% mark (8 lines)
I tried hard in this PR to minimize how often this happens.

src/components/markdown/markdown.tsx
```diff
@@ -15,13 +15,33 @@ import remarkRaw from 'rehype-raw';
...
+import mdxComponentMap, {
+  MarkdownComponentMap
+} from '../../utils/mdx-component-mapper';
 import indentNormalizer from '../../utils/indent-normalizer';
 import Notes from '../notes';
 import { ListItem } from '../../index';
 import { Appear } from '../appear';
 
-export const Markdown = React.forwardRef(
+type MdComponentProps = { [key: string]: any };
```
scottrippey, line 25: This file was the hardest to convert; I left a few items as `any` or `unknown`.

src/hooks/use-action-dispatcher.js
```diff
@@ -1,23 +0,0 @@
-import * as React from 'react';
```
scottrippey, line 1: (This file's behavior was inlined directly into `use-deck-state` for better type-safety)

scottrippey (commented):
Some things to note:

package.json
```diff
@@ -4,15 +4,14 @@
   "description": "ReactJS Powered Presentation Framework",
   "main": "lib/index.js",
```
scottrippey, line 5: Update: tested locally (via `yarn link`) with `spectacle-visual-editor` successfully!
image

src/index.ts
```diff
@@ -0,0 +1,53 @@
...
+  CodeSpan
+} from './components/typography';
+export {
+  Table,
+  TableCell,
+  TableRow,
+  TableHeader,
+  TableBody
+} from './components/table';
+export type { TableProps } from './components/table';
```
scottrippey, line 22: Rabbit hole here ...
to avoid compiler warnings, I had to use `export type` here...
to avoid prettier warnings, I had to upgrade to `prettier@2.x`
which requires arrow functions to use parenthesis (eg `(p) => p` not `p => p`)
which made tiny changes to about 100 files

So ...
sorry for such noise!

src/components/markdown/markdown.tsx
```diff
@@ -15,13 +14,33 @@ import remarkRaw from 'rehype-raw';
 import rehype2react from 'rehype-react';
 import { isValidElementType } from 'react-is';
```
carlos-kelly, line 15: Screen Shot 2021-11-18 at 3 36 52 PM

This works based on usage but I wonder if we should define this as a function that can take anything and returns a boolean?

src/components/markdown/markdown.tsx
carlos-kelly: It could be a serializable object, a React node, a function. But since it's a prop it could also be a falsy value. I think this makes sense.

src/components/slide/slide.tsx
```diff
@@ -1,27 +1,43 @@
...
 const noop = () => {};
 
-export const SlideContext = createContext(null);
+export const SlideContext = createContext<{
+  immediate: boolean;
+  slideId: SlideId;
+  isSlideActive: boolean;
+  activationThresholds: ActivationThresholds;
+  activeStepIndex: number;
+}>(null);
```
carlos-kelly, line 35: Does it make sense to make this an `export type`

ryan-roemer (commented):
Tried out `one-page.html` and `spectacle-cli` and everything works great on those ends!

Please open a version bump after we publish here: https://github.com/FormidableLabs/spectacle-cli/blob/main/package.json#L80

Thanks! 🎉

src/components/markdown/markdown.tsx
scottrippey: Added `@types/react-is`

src/components/slide/slide.tsx
scottrippey: Yes, makes total sense.  Extracted this to `SlideContextType`, and did the same for `DeckContextType`.

scottrippey:
@carlos-kelly Are the PR checks only configured to run against the `main` branch? Can we turn them on for the `v9` branch too?

ryan-roemer:
@scottrippey In this file: https://github.com/FormidableLabs/spectacle/blob/main/.github/workflows/ci.yml#L13-L18

Make this:

```yml
  pull_request:
    branches:
      - main
    paths:
      - '**'
      - '!docs/**'
```

this:

```
  pull_request:
    branches:
      - main
      # TODO(v9): Remove when targeting branch at `main` for real release.
      - v9
    paths:
      - '**'
      - '!docs/**'
```

src/index.ts
scottrippey: @carlos-kelly Great, another problem: eslint@8 dropped support for Node@10.
Our PR checks run against Node 10/12/14/16.  
Should we drop Node 10 support?  (It's, like, SUUUPER old)

ryan-roemer: @scottrippey -- Yes, please update to 12, 14, 16

scottrippey: Done!

carlos-kelly (approved):

PR #1098, opened 2022-03-02
Bug: Fix export mode

### LINKED ISSUE #1097: Export Mode is broken
### Prerequisites

**Feel free to delete this section if you have checked off all of the following.**

- [x] I've searched open [issues](https://www.github.com/FormidableLabs/spectacle/issues) to make sure I'm not opening a duplicate issue
- [x] I have read through the [docs](https://www.formidable.com/open-source/spectacle/docs) before asking a question
- [x] I am using the latest version of Spectacle

What version of Spectacle are you using? (can be found by running `npm list spectacle`)
`spectacle@9.0.0`
What version of React are you using? (can be found by running `npm list react`)
`react@17.0.2`
What browser are you using?
`chrome`
What machine are you on?
`macOS M1`

### Describe the Problem

Export mode is not working on main branch (and on 9.0.0).
For instance :
https://raw.githack.com/FormidableLabs/spectacle/main/examples/one-page.html?slideIndex=0&stepIndex=0&exportMode=true 

Error is : Cannot read properties of null (reading 'slideIndex')

**Expected behavior:** It should allow me to print my presentation as PDF.

**Actual behavior:** Blank page for now

ryan-roemer:
### Description

Our use of `const obj = merge(OBJECT, null)` produces `obj == null` which then causes problems upstream in export mode. This just adds some guardrails.

Now, at the higher level, I'm not totally following the reducer logic and how we get a null payload in there in export mode, but not other modes, so there might be a totally alternative, better solution than this.

Fixes #1097 

#### Type of Change

### How Has This Been Tested?

Put `one-page.html` into development mode and checked that local bundle works in export and normal modes.

### DIFF
--- a/src/hooks/use-deck-state.ts
+++ b/src/hooks/use-deck-state.ts
@@ -37,42 +37,37 @@ type ReducerActions =
   | { type: 'COMMIT_TRANSITION'; payload?: DeckView }
   | { type: 'CANCEL_TRANSITION'; payload?: undefined };
 
-// If payload is null, we don't want to merge because it nulls out the whole
-// object.
-const mergeView = (view: DeckView, payload: Partial<DeckView>) =>
-  payload ? merge(view, payload) : view;
-
 function deckReducer(state: DeckState, { type, payload = {} }: ReducerActions) {
   switch (type) {
     case 'INITIALIZE_TO':
       return {
-        activeView: mergeView(state.activeView, payload),
-        pendingView: mergeView(state.pendingView, payload),
+        activeView: merge(state.activeView, payload),
+        pendingView: merge(state.pendingView, payload),
         initialized: true
       };
     case 'SKIP_TO':
       return {
         ...state,
-        pendingView: mergeView(state.pendingView, payload)
+        pendingView: merge(state.pendingView, payload)
       };
     case 'STEP_FORWARD':
       return {
         ...state,
-        pendingView: mergeView(state.pendingView, {
+        pendingView: merge(state.pendingView, {
           stepIndex: state.pendingView.stepIndex + 1
         })
       };
     case 'STEP_BACKWARD':
       return {
         ...state,
-        pendingView: mergeView(state.pendingView, {
+        pendingView: merge(state.pendingView, {
           stepIndex: state.pendingView.stepIndex - 1
         })
       };
     case 'ADVANCE_SLIDE':
       return {
         ...state,
-        pendingView: mergeView(state.pendingView, {
+        pendingView: merge(state.pendingView, {
           stepIndex: 0,
           slideIndex: state.pendingView.slideIndex + 1
         })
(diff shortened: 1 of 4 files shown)


src/hooks/use-deck-state.ts
```diff
@@ -37,60 +37,65 @@ type ReducerActions =
   | { type: 'COMMIT_TRANSITION'; payload?: DeckView }
   | { type: 'CANCEL_TRANSITION'; payload?: undefined };
 
+// If payload is null, we don't want to merge because it nulls out the whole
+// object.
+const mergeView = (view: DeckView, payload: Partial<DeckView>) =>
+  payload ? merge(view, payload) : view;
+
 function deckReducer(state: DeckState, { type, payload = {} }: ReducerActions) {
   switch (type) {
```
scottrippey, line 46: Perhaps it would be more semantic, and easier to debug / remove later, if we just added another default like we do for `undefined`?
```
function deckReducer(state: DeckState, { type, payload = {} }: ReducerActions) {
  // Safeguard against nulls:
  if (!payload) {
    payload = {};
  }
  switch (type) {
```

scottrippey (approved):
Looks good.  
~I reviewed the code too, but couldn't find any place where a `null` could sneak in.  Strong types across the board, seems like everything is buckled down pretty tight.~
~The only scenario I could think of was if a user passed `<Deck initialState={null}`, but that would seem like a pretty obvious problem when trying to reproduce.  We could possibly add a runtime check there specifically, but this PR is a good improvement regardless.~
Update: looked deeper, found sneaky `null`

scottrippey (changes requested):
Oh wait, I actually found the bug ... and ironically it's something I broke when converting to TS.  Will post a fix asap.

src/hooks/use-location-sync.ts
```diff
@@ -72,9 +72,9 @@ export default function useLocationSync({
   ]);
 
   const syncLocation = useCallback(
-    (defaultState: DeckView) => {
+    (defaultState: DeckView): DeckView => {
       if (disableInteractivity) {
-        return null as unknown as DeckView;
+        return defaultState;
```
scottrippey, line 77: Before TS conversion ,this was just `return;`.  I mistakingly thought changing to `null` would be inconsequential.  
But returning `defaultState` is a better option.

src/hooks/use-deck-state.ts
```diff
@@ -37,60 +37,65 @@ type ReducerActions =
   | { type: 'COMMIT_TRANSITION'; payload?: DeckView }
   | { type: 'CANCEL_TRANSITION'; payload?: undefined };
 
+// If payload is null, we don't want to merge because it nulls out the whole
+// object.
+const mergeView = (view: DeckView, payload: Partial<DeckView>) =>
```
scottrippey, line 42: @ryan-roemer  With the null-bug fixed, what do you think about keeping these `mergeView` changes? 
IMO, null checks are like novocain; it numbs things but you end up biting yourself.  I can revert them if you want.

ryan-roemer: @scottrippey I'm down with whatever you feel is best -- I'm not that familiar with this code path / best practices and would love for you to just do what you think is best given the overall project and conventions..

src/hooks/use-location-sync.ts
ryan-roemer: I didn't trace through all the code paths, but I did have this thought of "shouldn't TS be catching this for us?" 😛

scottrippey: Yeah, but little did you know I was doing the classic Jedi-mind-trick on TypeScript! 
"null is a DeckView"

src/hooks/use-deck-state.ts
scottrippey: I rolled back this change, and verified that export mode is working correctly still 👍

scottrippey (approved):

ryan-roemer:
Thanks @scottrippey ! You want to merge and release or want me to do it?

scottrippey:
@ryan-roemer I'd prefer you, If you don't mind ... I don't have creds yet for NPM

ryan-roemer:
Released in `spectacle@9.0.1`

Thanks @scottrippey !

PR #1114, opened 2022-05-24
Fix: Newlines in markdown slides correctly render line breaks

buchansm:
### Description

Speculate was ignoring soft line returns for inline paragraph styles in slides created via markdown. Now the markdown slides accurately reflect each line break.

**Before:**
Before (1)

**After:**
After (1)

#### Type of Change

### How Has This Been Tested?

A unit test in `markdown.test.ts` checks that line breaks are correctly added for inline elements

### DIFF
--- a/src/components/markdown/markdown.test.tsx
+++ b/src/components/markdown/markdown.test.tsx
@@ -68,6 +68,20 @@ describe('<MarkdownSlide />', () => {
     );
     expect(wrapper.find('br')).toHaveLength(4);
   });
+
+  it('should generate line breaks for inline paragraph elements with carriage returns', () => {
+    const wrapper = mountInsideDeck(
+      <MarkdownSlide>{`One\r\n**Two**\r\n_Three_\r\n\`Four\`\r\n~~Five~~`}</MarkdownSlide>
+    );
+    expect(wrapper.find('br')).toHaveLength(4);
+  });
+
+  it('should generate line breaks for inline paragraph elements with mixed returns', () => {
+    const wrapper = mountInsideDeck(
+      <MarkdownSlide>{`One\n**Two**\r\n_Three_\n\`Four\`\r\n~~Five~~`}</MarkdownSlide>
+    );
+    expect(wrapper.find('br')).toHaveLength(4);
+  });
 });
 
 describe('<MarkdownSlideSet />', () => {
--- a/src/components/markdown/markdown.test.tsx
+++ b/src/components/markdown/markdown.test.tsx
@@ -55,6 +55,19 @@ describe('<MarkdownSlide />', () => {
     expect(wrapper.find('li').at(1).children().find('i')).toHaveLength(2);
     expect(wrapper.find('li').at(2).children()).toHaveLength(1);
   });
+
+  it('should generate line breaks for inline paragraph elements', () => {
+    const wrapper = mountInsideDeck(
+      <MarkdownSlide>{`
+        One
+        **Two**
+        _Three_
+        \`Four\`
+        ~~Five~~
+      `}</MarkdownSlide>
+    );
+    expect(wrapper.find('br')).toHaveLength(4);
+  });
 });
 
 describe('<MarkdownSlideSet />', () => {
--- a/src/components/markdown/markdown.tsx
+++ b/src/components/markdown/markdown.tsx
@@ -123,7 +123,7 @@ export const Markdown = forwardRef<HTMLDivElement, MarkdownProps>(
           // Replace \r\n and \n with <br /> for paragraphs
           const children =
             key === 'p'
-              ? props?.children?.map((child: any, i: number) => {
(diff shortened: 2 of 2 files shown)


src/components/markdown/markdown.tsx
```diff
@@ -119,9 +119,30 @@ export const Markdown = forwardRef<HTMLDivElement, MarkdownProps>(
...
       ).reduce((newMap, [key, Component]) => {
-        newMap[key] = (props: any) => (
-          <Component {...props} {...(componentProps || {})} />
-        );
+        newMap[key] = (props: any) => {
+          // Replace \n with <br /> for paragraphs
+          const children =
+            key === 'p'
+              ? props?.children?.map((child: any, i: number) => {
+                  if (child === '\n') return <br key={i} />;
```
ryan-roemer, line 127: Windows newlines?

buchansm: Yes, carriage returns weren't being rendered. The text string accurately reflects the line break inside the inspector, but the slide itself renders the text as inline.

carlos-kelly: I thought the Windows sequence was `\r\n`, which we might want to test for as well. It has the carriage return then new feed flags.

buchansm: Oh, I see. I didn't realize Windows had a different sequence.

ryan-roemer: Yep! And let's add test cases for win and mixed win+mac/lin

buchansm: I just put in a fix.

src/components/markdown/markdown.test.tsx
```diff
@@ -55,6 +55,19 @@ describe('<MarkdownSlide />', () => {
...
+      <MarkdownSlide>{`
+        One
+        **Two**
+        _Three_
+        \`Four\`
+        ~~Five~~
+      `}</MarkdownSlide>
+    );
+    expect(wrapper.find('br')).toHaveLength(4);
+  });
```
ryan-roemer, line 70: Could we add some win + mixed tests here? Thinking something like:

```suggestion

  it('should generate line breaks for inline paragraph elements with newline', () => {
    const wrapper = mountInsideDeck(
      <MarkdownSlide>{`
        One
        **Two**
        _Three_
        \`Four\`
        ~~Five~~
      `}</MarkdownSlide>
    );
    expect(wrapper.find('br')).toHaveLength(4);
  });

  it('should generate line breaks for inline paragraph elements with CR', () => {
    const wrapper = mountInsideDeck(
      <MarkdownSlide>{`One\r\n**Two**\r\n_Three_\r\n\`Four\`\r\n~~Five~~`}</MarkdownSlide>
    );
    expect(wrapper.find('br')).toHaveLength(4);
  });

  it('should generate line breaks for inline paragraph elements with mixed returns', () => {
    const wrapper = mountInsideDeck(
      <MarkdownSlide>{`One\n**Two**\r\n_Three_\n\`Four\`\r\n~~Five~~`}</MarkdownSlide>
    );
    expect(wrapper.find('br')).toHaveLength(4);
  });
```

src/components/markdown/markdown.tsx
```diff
@@ -119,9 +119,29 @@ export const Markdown = forwardRef<HTMLDivElement, MarkdownProps>(
...
+                    return lines.map((str, i) => (
+                      <React.Fragment key={i}>
+                        {str}
+                        {i !== lines.length - 1 && <br />}
+                      </React.Fragment>
+                    ));
+                  }
+                  return child;
+                })
+              : props.children;
```
ryan-roemer, line 138: Minor: We've got `props?.children` on L126 -- do we need the same safety guard here?

carlos-kelly: If it's falsey, typically null or undef, nothing will output unless it can be coerced like string "0" or "false".

ryan-roemer: I guess another way is asking -- can `props` ever be falsey, such that `.children` access will throw an error? If not, then we can remove the `?` from `props?.children` on L126

buchansm: Totally, I removed the safety guard from L126

src/components/markdown/markdown.test.tsx
buchansm: My bad 🤦‍♂️. I totally missed the part where you asked me to make tests for these. I thought you meant a regex test.

buchansm: I just added these tests.

src/components/markdown/markdown.tsx
carlos-kelly: It cannot when using JSX. So if you do `<Foo />` props will be `{}`. I'm sure there is a way to construct a component using a util function to make the props undefined, but that's not typical.

Issue #1106, opened 2022-04-19
Add unified template to presenter-mode deck


carlos-kelly:
### Description

Spectacle places the template on top of each slide individually. We should update it where the slide template overlays the deck but remains consistent when in presenter (default) mode.

This block renders the template for each slide currently:

```jsx
<TemplateWrapper style={wrapperOverrideStyle}>
  {templateElement}
</TemplateWrapper>
```

### Proposal

Let’s investigate lifting it out so template entities aren’t re-rendered each time per slide. This will also keep the template elements rendered and in place during slide transitions. Finally this will be key for the full effect of the PacMan animated progress component.

Issue #1105, opened 2022-04-19
Add animated PacMan Progress component


carlos-kelly:
### Description

[](https://github.com/FormidableLabs/spectacle/blob/v4.4.1/src/components/progress.js)Previous versions of Spectacle had a Pacman progress indicator instead of just progress dots. We should add a Pacman progress indicator component back to Spectacle.

### Proposal

The previous implementation can be seen

[https://github.com/FormidableLabs/spectacle/blob/v4.4.1/src/components/progress.js](https://github.com/FormidableLabs/spectacle/blob/v4.4.1/src/components/progress.js)

and can likely be heavily lifted for the new implementation. We should make it a separate component called `<AnimatedProgress />`

with similar color and size props as `<Progress />`

lenniezelk:
Hello, I'd like to work on this

carlos-kelly:
@lenniezelk Feel free to take a go at it! Thanks!

zubhav:
@lenniezelk Are you still working on this one? 🙂

lenniezelk:
Yes i am

Regards,

Lenny.

lenniezelk:
@carlos-kelly @zubhav  i'm studying the shared link https://github.com/FormidableLabs/spectacle/blob/v4.4.1/src/components/progress.js and I'm struggling to understand `Pacman.Base` and `Pacman.Body`
Kindly help me understand what is going on.

sauliuskerusauskas:
@lenniezelk - I hope you don't mind if I start taking a look at this one? 🙂 

I'm diving into a related improvement https://github.com/FormidableLabs/spectacle/issues/1106 that involves some refactoring and this component is going to be the perfect way for me to test my solution.

PR #1116, opened 2022-06-01
Feat: Persist deck template between slides

sauliuskerusauskas:
### Description

- Moved the deck-wide template up to deck level in default mode so that it persists between slides.
- Fixed deck level templates not being displayed in presenter, print and export modes.

Completes #1106 and paves the way for #1105.

To persist deck-wide templates between slides, I needed to move the rendering of the deck template up to deck level. If the slide has a slide-specific template however, we should not render the deck template, but render the slide template instead. For this, I needed to know whether the active slide had a slide-specific template at deck level. I added a `data-slide-has-template` attribute to the slides placeholder and extended the existing `useCollectSlides` hook, which we use to to surface slide ids to the deck, to also contain information about which slides have slide-specific templates. With the list of slides ids that have their own templates to hand, we can decide whether or not to show the deck template at deck level.

#### In default and presenter modes:
- If there is no slide specific template, the deck template will render at deck level and persist between slides.
- If there is a slide specific template, the deck template will not be diplayed and the slide template will be rendered at slide level.

Demo (with a transparent deck level template and blue slide template on the third slide):

https://user-images.githubusercontent.com/8139746/171488523-0b45687a-f2f0-41ce-8b0a-a0b2f31e6fd1.mov

#### In overview and print / export modes:
- For slides without a slide template, the deck template will be rendered at slide level.
- For slides with a slide template, the slide specific template will be rendered at slide level.

Screenshot 2022-06-01 at 17 08 19

#### Type of Change

No new functionality to add to docs - perhaps I should have categorised this one as a chore instead of a feature.

### DIFF
--- a/src/components/slide/slide.tsx
+++ b/src/components/slide/slide.tsx
@@ -24,6 +24,7 @@ import { ActivationThresholds, useCollectSteps } from '../../hooks/use-steps';
 import { GOTO_FINAL_STEP } from '../../hooks/use-deck-state';
 import { useSwipeable } from 'react-swipeable';
 import { SlideTransition } from '../transitions';
+import TemplateWrapper from '../template-wrapper';
 
 const noop = () => {};
 
@@ -36,6 +37,7 @@ export type SlideContextType = {
 };
 
 export const SlideContext = createContext<SlideContextType>(null as any);
+SlideContext.displayName = 'SlideContext';
 
 type SlideContainerProps = BackgroundProps &
   ColorProps & { backgroundOpacity: number };
@@ -73,15 +75,6 @@ const SlideWrapper = styled.div<ColorProps & SpaceProps>(
   `
 );
 
-const TemplateWrapper = styled.div`
-  position: absolute;
-  top: 0;
-  left: 0;
-  right: 0;
-  bottom: 0;
-  pointer-events: none;
-`;
-
 export const AnimatedDiv = styled(animated.div)`
   width: 100%;
   height: 100%;
@@ -115,7 +108,8 @@ const Slide = (props: SlideProps): JSX.Element => {
     throw new Error(`Slide components may not be nested within each other.`);
   }
 
-  const { slideId, placeholder } = useSlide(userProvidedId);
+  const hasTemplate = Boolean(template);
+  const { slideId, placeholder } = useSlide(hasTemplate, userProvidedId);
   const { setStepContainer, activationThresholds, finalStepIndex } =
     useCollectSteps();
   const {
@@ -312,20 +306,24 @@ const Slide = (props: SlideProps): JSX.Element => {
     };
   }, [wrapperOverrideStyle, theme, padding]);
 
-  const templateFn =
-    typeof template === 'function'
-      ? template
-      : typeof deckTemplate === 'function'
-      ? deckTemplate
-      : null;
+  const slideTemplateElement: ReactNode = useMemo(() => {
+    return typeof template === 'function'
+      ? template?.({
+          slideNumber: activeView.slideIndex + 1,
+          numberOfSlides: slideCount
+        })
+      : template;
+  }, [template, activeView.slideIndex, slideCount]);
 
-  const templateElement: ReactNode =
-    templateFn?.({
-      slideNumber: activeView.slideIndex + 1,
-      numberOfSlides: slideCount
-    }) ||
-    template ||
-    deckTemplate;
+  const deckTemplateElement: ReactNode = useMemo(() => {
+    if (!inOverviewMode) return null;
+    return typeof deckTemplate === 'function'
+      ? deckTemplate?.({
+          slideNumber: activeView.slideIndex + 1,
+          numberOfSlides: slideCount
+        })
+      : deckTemplate;
+  }, [inOverviewMode, deckTemplate, activeView.slideIndex, slideCount]);
 
   const swipeHandler = useSwipeable({
     onSwiped: (eventData) => onMobileSlide(eventData)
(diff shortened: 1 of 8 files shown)


sauliuskerusauskas:
When we move to a slide with a template, the template can either be inside or outside of the slide coming in:

Inside:

https://user-images.githubusercontent.com/8139746/171488140-325808db-ecba-4a18-a470-b68ddc5da085.mov

Outside:

https://user-images.githubusercontent.com/8139746/171487976-f828099c-6dbf-4e55-94fb-9d99dddf6432.mov

I've gone with having the template inside but could easily tweak it to be outside.

src/components/slide/slide.tsx
```diff
@@ -154,7 +150,6 @@ const Slide = (props: SlideProps): JSX.Element => {
     return result;
   }, [slideTransition, transition]);
 
-  const inOverviewMode = Object.entries(frameOverrideStyle).length > 0;
```
sauliuskerusauskas, line 157: `frameOverrideStyle` has entires in both overview and print / export modes, so `isOverviewMode` was `isInOverviewOrPrintMode`. To distinguish between overview and print modes I've added booleans for both to `DeckContext` which I'm now using instead.

src/hooks/use-slides.tsx
```diff
@@ -12,6 +12,8 @@ export function useCollectSlides() {
   const [initialized, setInitialized] = useState(false);
   const [slideContainer, setSlideContainer] = useState<HTMLElement | null>();
   const [slideIds, setSlideIds] = useState<SlideId[]>([]);
+  const [slideIdsOfSlidesWithTemplates, setSlideIdsOfSlidesWithTemplates] =
+    useState<SlideId[]>([]);
```
gksander, line 16: It looks like the primary use-case for this state is to check whether or not a slide ID has a template, and we don't care so much about order of IDs etc.. In this case, I recommend we use a `Set` instead of an array, so that our "contains" lookup is `O(1)` complexity instead of `O(n)`.

gksander (commented):
This is looking awesome! Two 🤏 suggestions here:

- Can we remove the `template.tsx` file? It's empty, looks like maybe it was accidentally added.
- Left a comment in the review about using a Set instead of array for storing IDs for slides that have their own templates.

gksander:
One thing I noticed with this branch is that export mode now shows some extra UI, like the fullscreen indicator and steps:

image

@carlos-kelly is export mode supposed to have those UI controls compressed/hidden?

src/hooks/use-slides.tsx
sauliuskerusauskas: Good idea, have updated to be a Set in: https://github.com/FormidableLabs/spectacle/pull/1116/commits/7a336940c61e50bda402852eaa5c790a4cdb1a38

sauliuskerusauskas:
> One thing I noticed with this branch is that export mode now shows some extra UI, like the fullscreen indicator and steps:
> 
> @carlos-kelly is export mode supposed to have those UI controls compressed/hidden?

@gksander My understanding from Carlos' reply in the spectacle slack channel a couple days ago was that the template should be shown in export mode but both the full screen toggle and the progress bars are hidden with `@media print` queries as soon as we try to print. 

I can reason about this by for-seeing elements that might be added to the template (like a logo), that unlike the full screen toggle or progress bar, we might still want to display when printing, so it makes sense to show the template rather than hide it, and then decide whats hidden or not on a per-component basis.

For the full screen and progress components specifically, if we think the slight discrepancy before and after the print modal is opened might be confusing, I can suggest hiding both by having them return `null` in print and export modes as well as hiding both with print targeted media queries?

carlos-kelly:
@gksander @sauliuskerusauskas Correct. Seen here:
https://github.com/FormidableLabs/spectacle/blob/main/src/components/fullscreen.tsx#L14

Export Mode !== Print Mode

Print Mode drops the background so printing doesn’t waste ink and can work on B&W printers
Export mode is for flattening the deck so it can be saved as a PDF

Some browsers combine the two functions, but we explicitly go an extra step in making `print` mode as optimized as possible for a paper format. With that said, I do not feel we need to hide controls on export mode.

sauliuskerusauskas:
@carlos-kelly Yep, all makes sense. I understand print mode to be export mode + some extra colour swaps in the theme to optimise for B&W printing.

To make sure we're all on the same page, prior to this change the controls did not display on neither print, nor export modes. After this fix, the controls will display on both:

Export mode:
Screenshot 2022-06-07 at 10 52 48

Print mode:
Screenshot 2022-06-07 at 10 53 09

up until the user launches the print modal (to either print in print mode, or to export as pdf in export mode). At this point the print stylesheet kicks in and means the controls don't end up in the output of either, neither the print nor the saved .pdf.

In both modes, what you see is not necessarily what you get up until you launch the print modal, which is why I suggested we could hide the controls in both modes as they are outputted in neither, in case that led to any confusion. But also think it's fine as it is, so long as we're clear that after this change we'll be showing the controls on both modes and outputting the controls in neither, in case I miscommunicated about them being displayed or in the output of one mode and not the other.

src/hooks/use-slides.tsx
```diff
@@ -20,23 +22,40 @@ export function useCollectSlides() {
     ) as unknown as Iterable<HTMLElement>;
 
     const nextSlideIds: SlideId[] = [];
+    const nextSlideIdsOfSlidesWithTemplates: Set<SlideId> = new Set();
     for (const placeholderNode of slides) {
-      const { slideId } = placeholderNode.dataset;
+      const { slideId, slideHasTemplate } = placeholderNode.dataset;
       nextSlideIds.push(slideId!);
+      if (slideHasTemplate === 'true') {
+        nextSlideIdsOfSlidesWithTemplates.add(slideId!);
```
gksander, line 30: I know there's an existing `slideId!` (with the postfix `!` annotation), so it makes sense to just do the same here. In general, I feel pretty strongly against the `!` annotation though, because it limits the utility of TS (we're saying we _know_ the value is defined, even if we haven't actually checked that it is). 

Perhaps in a future PR, we can address that issue and clean up both of these `slideId!` instances, just by doing a check that `slideId !== undefined`.

gksander (approved):
LGTM, great work! 🚀

PR #1119, opened 2022-06-08
Upgrade to React 18

gksander:
### Description

It's cool to be _cool_. This PR upgrades our dependencies to React 18 so that we can be cool. This upgrade opened a bit of a can of worms, and I tried my darndest to split this out but every piece seemed to be tied together:

- Upgrade our peer dependencies to support React 18.
- Upgrade `@types/react` etc to 18, which then forced me to fix some TS issues that popped up.
- Enzyme does not support React 18 (and Enzyme is 💀 ) so dropped Enzyme in favor of RTL.
- Had to rewrite tests into RTL accordingly, and decided to drop snapshots in favor of actual component testing (otherwise all of the snapshots would have huge diffs, and I believe RTL tests provide more value than the snapshot tests).
- Updated our examples to also use React 18 to ensure compatibility.

#### Type of Change

- [ ] Regular dependency maintenance.

### How Has This Been Tested?

Tests pass. Sample apps all still run.


examples/js/index.js
```diff
@@ -264,4 +264,5 @@ const Presentation = () => (
   </Deck>
 );
 
-ReactDOM.render(<Presentation />, document.getElementById('root'));
+const root = createRoot(document.getElementById('root'));
+root.render(<Presentation />);
```
gksander, line 268: React 18 change to mounting.

package.json
```diff
@@ -93,24 +92,23 @@
     "@babel/preset-env": "^7.12.7",
     "@babel/preset-react": "^7.16.7",
     "@babel/preset-typescript": "^7.16.0",
-    "@testing-library/react-hooks": "^3.7.0",
```
gksander, line 96: Hooks testing has been absorbed into `@testing-library/react`, so removing this dep.

package.json
```diff
@@ -119,17 +117,21 @@
...
+    "react-dom": "^18.1.0",
     "rimraf": "^3.0.0",
     "ts-jest": "^27.0.7",
     "typescript": "^4.5.2",
     "webpack": "^5.68.0",
     "webpack-cli": "^4.9.2",
     "webpack-dev-server": "^4.7.4"
+  },
+  "resolutions": {
+    "@types/react": "^18.0.12"
```
gksander, line 135: Conflicting types in styled-system were fighting me, so I cheated and used a resolution – but let me know if you have any other ideas. Can remove if styled-system eventually updates their types dependency to something compatible with ours.

package.json
```diff
@@ -75,15 +74,15 @@
...
     "ulid": "^2.3.0",
     "unified": "^9.0.0",
     "unist-util-visit": "^2.0.3",
     "use-resize-observer": "^6.1.0"
   },
   "peerDependencies": {
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2"
+    "react": ">=17.0.2",
+    "react-dom": ">=17.0.2"
```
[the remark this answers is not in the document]
carlos-kelly: Correct, but it would pin to whatever is in the lock file.

package.json
carlos-kelly: I've had to do this before.

src/components/slide/slide.tsx
```diff
@@ -335,6 +336,7 @@ const Slide = (props: SlideProps): JSX.Element => {
           ReactDOM.createPortal(
             <AnimatedDiv
               ref={setStepContainer}
+              // @ts-expect-error Events are not typed quite tightly enough (yet)
```
carlos-kelly, line 339: 😭

carlos-kelly (approved):

PR #1125, opened 2022-06-10
fix: only the first page showing in export and print modes in Firefox

### LINKED ISSUE #1077: Bug: Export Mode doesn't work in Firefox.
### Question

Hello 👋 

I'm building some slides right now and wanted to look an a PDF export but oddly only the first Slide is visible but the PDF has the full 121 slides as pages but they are all white. I looked into similiar issues where ppl said imported slides have some issue and moved all CustomSlides which were imported into the `App.tsx` but that did produce a different result. I used Firefox to generate the PDF.

TL;DR: Chrome is working Firefox not.

Is Firefox not supported or is this a bug?

### Background Info/Attempts

I used:
* Firefox 94.0.1 (64Bit) as a browser
* react ^17.0.2
* spectacle ^8.4.1
* typescript ^4.1.2

Here a screen of the PDF generation preview:
Bildschirmfoto 2021-11-26 um 12 46 09

You can see, only the first slide gets into the PDF and even that slides is a little bit off.

After that i tried this simple example:
```
const theme = {
    colors: {
        primary: '#333333',
        secondary: '#39bbcf',
        tertiary: '#efefef',
        quaternary: '#efefef',
        grey: '#ddd',
        red: '#c92a2a'
    },
    fonts: {
        primary: 'Ubuntu',
    }
};

function App() {
  return (
      <Deck theme={theme} transition={{from: {}, leave: {}, enter: {}}}>
          <Slide backgroundColor="primary">
              <FlexBox height="100%">
                  <Heading size={3} lineHeight={1} color="secondary">
                      Slide 1
                  </Heading>
              </FlexBox>
          </Slide>
          <Slide backgroundColor="primary">
              <FlexBox height="100%">
                  <Heading size={3} lineHeight={1} color="secondary">
                      Slide 2
                  </Heading>
              </FlexBox>
          </Slide>
      </Deck>
  );
}
```
With this i get the same result, only first slide and still off:
Bildschirmfoto 2021-11-26 um 12 51 54

I'll downloaded chrome and tried all steps above backwards and everything worked:
Bildschirmfoto 2021-11-26 um 13 03 04

sauliuskerusauskas:
### Description

Before, only the first slide was displaying when performing the print / export to pdf actions in print and export modes in Firefox. After this change, all of the slides should display.

Fixes #1077 

Before:

https://user-images.githubusercontent.com/8139746/173054235-20e4f158-f674-4467-8e8a-1acbdb91006c.mov

And after:

https://user-images.githubusercontent.com/8139746/173054251-b2948960-3184-422d-a6d2-75763b78ab24.mov

#### Type of Change

### How Has This Been Tested?

Tested before and after in Firefox and across all the modes in Chrome to make sure the change is O.K.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 - Fix deck level templates not displayed in presenter, print and export modes
 - Persist deck template between slides in default and presenter modes [#1106](https://github.com/FormidableLabs/spectacle/issues/1106)
+- Update to react 18
+- Replace enzyme tests with react-testing-library tests
+- Fix only the first page showing in print and export modes in Firefox
 
 ## 9.1.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,8 @@
 
 - Fix deck level templates not displayed in presenter, print and export modes
 - Persist deck template between slides in default and presenter modes [#1106](https://github.com/FormidableLabs/spectacle/issues/1106)
-- Update to react 18
-- Replace enzyme tests with react-testing-library tests
-- Fix only the first page showing in print and export modes in Firefox
+- Upgrade dependencies to be React 18-friendly, removing Enzyme in favor of RTL via [#1119](https://github.com/FormidableLabs/spectacle/pull/1119)
+- Fix only the first page showing in print and export modes in Firefox [#1077](https://github.com/FormidableLabs/spectacle/issues/1077)
 
 ## 9.1.1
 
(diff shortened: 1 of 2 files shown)


gksander (approved):
Nice find! Interesting to me that the Firefox PDF export doesn't fit the page to the slides like Chromium. Also, Safari's print→PDF is pretty wonked too. In the future, we might try to investigate a x-browser compatible way to generate these previews (something that's always pretty tricky).

PR #1126, opened 2022-06-10
chore: remove broken size badge from README

sauliuskerusauskas:
### Description

In the `README.md` the un-compressed size badge appears with a 'critical error' instead of the size.

Screenshot 2022-06-10 at 12 27 11

It appears that providing an absolute URL to the `badge-size` service is currently broken for getting the non gzipped size while gzipped sizes still work. There is an an outstanding in the `badge-size` project: https://github.com/ngryman/badge-size/issues/94 but it has been open for a little while now.

I noticed other repos of ours only display the gzipped size. Removing the un-compressed size from  the `README.md` so the 'critical error' doesn't unnecessarily worry anyone.


gksander (approved):
Nice, I like it. We'll keep an eye on that unpkg issues and perhaps re-add this once that issue is resolved.

PR #1131, opened 2022-06-30
`use-broadcast-channel` fix

### LINKED ISSUE #1130: BroadcastChannel gets closed before being used
https://github.com/FormidableLabs/spectacle/blob/00c337f62c22d0e8b1ffa1ec0ab45286fe760caa/src/hooks/use-broadcast-channel.ts#L25-L32

With React 18, in Strict Mode the effects are firing twice, which makes this code break as the channel is closed and then never opened again because the name is the same. Probably a better approach here would be to use a ref instead of state as it will also cause fewer re-renders.

I can make a pull request if needed.

gksander:
### Description

This PR tweaks `use-broadcast-channel` to use refs instead of state. With React 18 strict mode, effects run twice – and our existing effects setup was not setup to handle that. This PR stabilizes things a bit, and uses refs instead of state (since we don't need these pieces of data to trigger UI rerenders).

Fixes #1130 

#### Type of Change

### How Has This Been Tested?

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] **I've tested this change manually in presenter mode.**

### DIFF
--- a/src/hooks/use-broadcast-channel.ts
+++ b/src/hooks/use-broadcast-channel.ts
@@ -1,4 +1,4 @@
-import { useState, useEffect, useCallback, useRef } from 'react';
+import { useCallback, useEffect, useRef } from 'react';
 import { ulid } from 'ulid';
 import { BroadcastChannel as BroadcastChannelPolyfill } from 'broadcast-channel';
 import { DeckView } from './use-deck-state';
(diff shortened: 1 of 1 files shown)


src/hooks/use-broadcast-channel.ts
```diff
@@ -17,19 +17,16 @@ export default function useBroadcastChannel(
...
-    () => new BroadcastChannel(channelName)
-  );
+  const broadcasterId = useRef(() => ulid());
+  const channel = useRef<BroadcastChannel>();
 
   useEffect(() => {
-    if (channel.name !== channelName) {
-      setChannel(() => new BroadcastChannel(channelName));
-    }
+    channel.current = new BroadcastChannel(channelName);
```
gksander, line 24: I don't think we really care about channel name check. Any time `channelName` changes, let's tear down the old channel and create a new one.

src/hooks/use-broadcast-channel.ts
```diff
@@ -17,19 +17,16 @@ export default function useBroadcastChannel(
   onMessage: MessageCallback = noop,
   deps = []
 ) {
-  const [broadcasterId] = useState(() => ulid());
-  const [channel, setChannel] = useState(
-    () => new BroadcastChannel(channelName)
-  );
+  const broadcasterId = useRef(() => ulid());
```
carlos-kelly, line 20: Do we need `ulid` with `useId()` in React 18?

gksander: We have a React peer dep of `>= 17.0.2` (or something like that), so I think using a React-18 specific API would break that, no?

carlos-kelly: Ah yes, not worth adjusting that here.

PR #1132, opened 2022-06-30
Vite fix

### LINKED ISSUE #1118: Issues with Prism in Vite/ESBuild-powered projects
### Describe the Problem

When attempting to use Spectacle in a Vite-powered React project, the presentation fails to run with the following error:

image

It appears we're running into some issues with Prism/prism-react-renderer. We should investigate and resolve this issue so that Spectacle works out of the box with Vite/other ESM-powered build tools.

gksander:
### Description

**Waiting for #1131 to go through, since this is branched from that.**

Upgrades `react-syntax-highlighter`, which appears to not affect our functionality – but upgrades some nested dependencies that were previously throwing errors in a vanilla vite setup. With this upgrade, spectacle now seems to run out of the box with Vite 🥳 

### Sample

(This only works after this is shipped, but you can follow along by running `yarn pack` in the spectacle repo and install from a tgz prior to this getting shipped)

Run 

```sh
yarn create vite
```

and select the React + Typescript starter. Then

```sh
yarn add spectacle
```

and in `App.tsx` drop in something simple like:

```tsx
import { Deck, Slide, Text, SlideLayout } from "spectacle";

function App() {
  return (
    <Deck>
      <SlideLayout.List
        title="Things I like"
        items={["Pokemon", "Getting swol", "Comedy"]}
      />
    </Deck>
  );
}

export default App;
```

and fire up the app with `yarn dev`. Should work out of the box.

Fixes #1118 .


carlos-kelly (approved):
I think this branch absorbed some of the changes for broadcast channel.

PR #1138, opened 2022-07-05
Added base CLI with commander, tsconfig

carlos-kelly:
### Description

Adds a base CLI tool with commander and chalk. Still a work in progress, but any eyes over the tsconfig and setup is great.

Current Console UX showing error handling:

Screen Shot 2022-07-05 at 4 56 18 PM

#### Type of Change


ryan-roemer:
Neat idea! Is this just meant to be for development or adding to real bin for package? 

If former, we should be adding just devDeps.

If latter then we likely need lots more prod deps (which is the reason we originally separated the boiler plate to a different package) and also note that boilerplate places a non script named “spectacle-cli” in addition to “spectacle-boilerplate” that will conflict with this (see https://github.com/FormidableLabs/spectacle-cli)

carlos-kelly:
@ryan-roemer Yup, it's a replacement but I think also exploring slimming down in features. The previous CLI had a lot of complexity when I think the goal was just "bootstrap a project". Still in exploration but the goal is unifying the codebases as they always got out of sync.

PR #1139, opened 2022-07-07
Conversion to `pnpm` workspaces 

gksander:
## PNPM Workspaces 🎈 

This PR moves the `spectacle` repo into a pnpm workspaces setup so that we can migrate the CLI (and other utilities) into this repo. We'll follow up with the addition of `wireit`, which will make the package scripts more declarative and performant. To start, I just got all of the scripts _working_ – we'll optimize once we get to adding `wireit`.

I'll litter this PR with comments to point out things I ran into.


.github/workflows/ci.yml
```diff
@@ -25,32 +24,36 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x]
```
gksander, line 28: Push back on this if you want. 12 is EOL, not sure there's much sense in supporting it anymore (but I could be convinced otherwise).

examples/js/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "../../.babelrc"
+}
```
gksander, line 3: Babel fought me. Unless I had a `.babelrc` in each repo, I couldn't get `babel-loader` etc working properly, so I opted for just adding `.babelrc` wherever needed and just extending from the base config.

examples/js/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "spectacle-example-js",
```
gksander, line 2: Each of the examples is now a part of the workspace and have their own scripts/deps (most of the deps are just shared so no extra baggage, just have to declare them here). This allows us to run scripts in all of the examples at once, etc. - and allows us to link to the workspace spectacle code so we can update spectacle core code and have it reflected live-time in our examples (without any linking etc)

examples/js/webpack.config.js
```diff
@@ -18,11 +18,5 @@ module.exports = {
       title: 'Spectacle JavaScript Demo',
       template: `./index.html`
     })
-  ],
-  resolve: {
-    ...base.resolve,
-    alias: {
-      spectacle: path.resolve(__dirname, '../../src')
-    }
-  }
```
gksander, line 27: No longer need to resolve here, we'll just use pnpm workspace settings (and we'll use the actual build files instead of src, which gives a more accurate depiction of real-world use)

packages/spectacle/src/components/code-pane.tsx
```diff
@@ -184,6 +184,7 @@ const CodePane = forwardRef<HTMLDivElement, CodePaneProps>(
     );
   }
 );
+CodePane.displayName = 'CodePane';
```
gksander, line 187: ESLint yelled at me. I think pnpm installed a newer version of `eslint-plugin-react`.

packages/spectacle/src/components/slide-layout.tsx
```diff
@@ -66,6 +66,7 @@ const List = ({
           {title}
         </Heading>
       ) : null}
+      {/* @ts-ignore TODO: Resolve this in follow-up */}
       <List {...listProps}>
```
gksander, line 70: TS yelling about spreading `listProps` here, need to just do a little bit of TS shenanigans.

packages/spectacle/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+  "include": ["src/**/*"],
+  "compilerOptions": {
+    "baseUrl": ".",
+    "declaration": true,
+    "declarationMap": true,
+    "sourceMap": true,
+    "outDir": "./lib",
+    "paths": {
+      "csstype": ["node_modules/csstype"]
+    }
```
gksander, line 12: [See here](https://github.com/microsoft/TypeScript/issues/42873)

tsconfig.json
```diff
@@ -1,14 +1,10 @@
 {
-  "include": [
-    "src/**/*", 
-    "examples/typescript/**/*"
-  ],
```
gksander, line 5: Moved some of these settings into the `packages/spectacle/tsconfig.json` config

package.json
```diff
@@ -1,137 +1,40 @@
 {
-  "name": "spectacle",
```
gksander, line 2: New root `package.json` file – mostly just here for some repo-wide dev deps (like ESLint, prettier, etc.) and to run some root scripts. Do not need any npm-specific package details, since we don't release from this `package.json` – the file details will like in `packages/**/package.json` files.

gksander (commented):
Leaving some comments.

.github/workflows/ci.yml
ryan-roemer: We should be doing 14, 16, 18

gksander: This comment is dated, I added 18 to this matrix after leaving the comment 👍

examples/js/.babelrc
carlos-kelly: That's typical. With our Nx monorepo we needed separate ones too.

packages/spectacle/src/components/code-pane.tsx
carlos-kelly: Ah yeah this makes sense as `CodePane` is declared as a `forwardRef` anon function.

carlos-kelly (approved):

PR #1142, opened 2022-07-08
Add Base CLI package

carlos-kelly:
### Description

Add base CLI package using Commander, Chalk to generate `jsx`, `mdx`, `tsx`, and `onepage`.

- One-page fully works

**Mac**
177423999-bf5b8827-7128-4bca-a840-e3e69c613332

**Windows**
image


packages/cli/src/templates/index.ts
```diff
@@ -0,0 +1,27 @@
...
+type FileOptions = {
+  snakeCaseName: string;
+  name: string;
+};
+
+export const writeBaseWebpackProjectFiles = async ({
+  snakeCaseName,
+  name
+}: FileOptions) => {
+  if (existsSync(snakeCaseName)) {
```
gksander, line 15: Idk why this threw me off a bit, but this is basically an implicit relative path, right? Wonder if it'd be more readable to have something like:

```ts
const outPath = path.resolve(process.cwd(), snakeCaseName);
```

and then use that throughout this fn.

gksander (approved):
LGTM. Couple minor nits, but nothing mission-critical.

PR #1145, opened 2022-07-14
Added file writers to complete JSX/TSX deck generation.

carlos-kelly:
### Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

#### Type of Change

### How Has This Been Tested?

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My code follows the style guidelines of this project (I have run `yarn format`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes (I have run `yarn test`)
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -5,16 +5,20 @@ import { Command } from 'commander';
 import cliSpinners from 'cli-spinners';
 import logUpdate from 'log-update';
 import {
-  writeBaseWebpackProjectFiles,
+  FileOptions,
+  writeWebpackProjectFiles,
   writeOnePageHTMLFile
-} from './templates';
+} from './templates/file-writers';
 
 type CLIOptions = {
   type: 'tsx' | 'jsx' | 'mdx' | 'onepage';
   name: string;
+  lang?: string;
+  port?: number;
 };
 
 let progressInterval: NodeJS.Timer;
+const SPECTACLE_VERSION = '9.2.1';
 const log = console.log;
 const program = new Command();
 const printConsoleError = (message: string) =>
@@ -41,11 +45,12 @@ const main = async () => {
       'deck source type (choices: "tsx", "jsx", "mdx", "onepage")'
     )
     .requiredOption('-n, --name [name]', 'name of presentation')
+    .option('-l, --lang [lang]', 'language code for generated HTML document')
+    .option('-p, --port [port]', 'default port for webpack dev server')
     .parse(process.argv);
 
   let i = 0;
-  const { type, name } = program.opts<CLIOptions>();
-  const snakeCaseName = name.toLowerCase().replace(/([^a-z0-9]+)/gi, '-');
+  const { type, name, lang = 'en', port = 3000 } = program.opts<CLIOptions>();
 
   progressInterval = setInterval(() => {
     const { frames } = cliSpinners.aesthetic;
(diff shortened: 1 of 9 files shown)


packages/cli/src/cli.ts
```diff
@@ -5,16 +5,20 @@ import { Command } from 'commander';
...
 
 type CLIOptions = {
   type: 'tsx' | 'jsx' | 'mdx' | 'onepage';
   name: string;
+  lang?: string;
+  port?: number;
 };
 
 let progressInterval: NodeJS.Timer;
+const SPECTACLE_VERSION = '9.2.1';
```
carlos-kelly, line 21: I don't like this, but importing the version number from the core library `package.json` introduces other cross complexities. Open to ideas.

packages/cli/src/templates/index.ts
```diff
@@ -1,29 +1,46 @@
...
+      <FlexBox height="100%">
+        <Heading fontSize="h2">Made with</Heading>
+        <SpectacleLogo size={300} />
+      </FlexBox>
+    </Slide>
+  </Deck>
+);
+
+createRoot(document.getElementById('app')${
+    options.usesTypeScript ? '!' : ''
```
carlos-kelly, line 44: TypeScript needs `!` since that element by ID could be null. We're confident here to force unwrap since the element ID is not dynamic and the HTML is included.

packages/cli/src/cli.ts
gksander: If there are added complexities, this seems fine. Maybe we can just write a little node-script to automate updating this on release or something.

packages/cli/src/cli.ts
```diff
@@ -41,11 +45,12 @@ const main = async () => {
       'deck source type (choices: "tsx", "jsx", "mdx", "onepage")'
     )
     .requiredOption('-n, --name [name]', 'name of presentation')
+    .option('-l, --lang [lang]', 'language code for generated HTML document')
+    .option('-p, --port [port]', 'default port for webpack dev server')
```
scottrippey, line 49: Maybe the description should mention the default values?

ryan-roemer:
It's going to add build time, but we should add "generate, install, and build" CI steps in a separate workflow in parallel for each CLI generate scenario (what the existing CLI does)

packages/cli/src/templates/package.ts
```diff
@@ -0,0 +1,45 @@
...
+  `{
+  "name": "${options.name}",
+  "private": true,
+  "scripts": {
+    "start": "webpack-dev-server --hot --config ./webpack.config.js",
+    "clean": "rimraf dist",
+    "build": "webpack --config ./webpack.config.js --mode production"
+  },
+  "dependencies": {
+    "spectacle": "${options.spectacleVersion}",
```
scottrippey, line 17: Super minor, but do you think including `^` in the version number feels more appropriate?

carlos-kelly:
@ryan-roemer Yeah we'd definitely want to continue ensuring the CI works for all of the 3 flows. I'm open to ideas on verifying the `one-page` file generated actually works. We'd need a generous timeout I think.

scottrippey (approved):
Looks really great!  Love the separation of the files and functions.

packages/cli/src/cli.ts
carlos-kelly: Good idea!

gksander (approved):
DOPE

packages/cli/package.json
```diff
@@ -27,6 +27,7 @@
   },
   "peerDependencies": {},
   "devDependencies": {
+    "spectacle": "workspace:spectacle@*",
```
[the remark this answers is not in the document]
carlos-kelly: If it's a runtime it will be included in the entire bin when we build this cli as runnable. We likely need to do a script to extract out the version number from the other package.json rather than some kind of cross-package dependency. @gksander and I were pondering this, but I'm really mixed if typing the version is easier. We do that on one-page today.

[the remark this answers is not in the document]
scottrippey: Yeah I'm mixed too ... hard-coding it seems easy, but not as easy to maintain.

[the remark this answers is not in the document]
scottrippey: I think we do need to change this before publishing, because currently trying to run the CLI (outside this repo) I think will fail.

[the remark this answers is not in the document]
carlos-kelly: Yup! the CLI is not ready to be published yet. We still need to build the interactive prompts. Just flags are working.

PR #1151, opened 2022-07-20
Script/Publish cleanup

gksander:
This PR adds a handful of minor things to our scripts/publish setup:

- Adds a `nuke` script to nuke everything, seems like it's sometimes needed to get a fresh start with a `pnpm install`.
- Move webpack deps into example packages.
- Move CHANGELOG into spectacle package.
- On prepublish, copy our root README into spectacle package so that it's available on NPM.

Bonus addition: vsc local settings to ignore built files, so file searching is a lot better. JB doesn't make this as easy, so I unfortunately don't have an equivalent for the JB folks.

image


packages/spectacle/package.json
```diff
@@ -7,10 +7,7 @@
   "files": [
     "lib/",
     "es/",
-    "dist/",
-    "examples/",
```
ryan-roemer, line 11: Note that in the past we published `examples` so that it could be used upstream as a dep of `spectacle-ci` for generation. We don't have to continue doing this, but just for some historical context.

gksander: That's good context! Hopefully with the monorepo setup, we won't have to ship those. (But also, the examples are no longer under the spectacle package, so I originally removed this because they wouldn't actually be included in the bundle as it currently stands.)

.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "search.exclude": {
+    "**/node_modules": true,
+    "**/lib": true,
+    "**/es": true,
+    "**/dist": true
+  }
+}
```
carlos-kelly, line 8: I think a vscode file made it through

package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "spectacle-monorepo",
   "scripts": {
-    "start:js": "pnpm run --filter spectacle-example-js start",
-    "start:ts": "pnpm run --filter spectacle-example-ts start",
-    "start:md": "pnpm run --filter spectacle-example-md start",
+    "start:js": "pnpm run build:core && pnpm run --filter spectacle-example-js start",
+    "start:ts": "pnpm run build:core && pnpm run --filter spectacle-example-ts start",
+    "start:md": "pnpm run build:core && pnpm run --filter spectacle-example-md start",
```
carlos-kelly, line 6: should we watch these that way changes to the library update in the demo apps?

gksander: Yeah, I think that's the right move. I had that same vision, but just didn't get it squeezed into this PR. Happy to try to do that, or we can tackle as a follow-up.

carlos-kelly: To have a watch be a task doesn't it need to use some concurrent helper otherwise since the process never sees an exit it will never execute the right hand of the &&?

carlos-kelly (approved):

package.json
gksander: Yeah, I think we'll want to pull in concurrently for this. Roemer setup a similar script in Victory using concurrently, we can borrow from that!

PR #1166, opened 2022-07-29
CLI build

gksander:
## What?

Adds a `prepublishOnly` script to run the build process before publish, and update `main`/`files` to point to the `bin`'s cli.js file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,7 @@ dist
 lib
 es
 bin
+
+# Pack-ing artifacts
+packages/**/package
+**/*.tgz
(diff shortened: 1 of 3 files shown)


carlos-kelly (approved):

ryan-roemer:
One thing I haven’t looked up is doing npm tags so we don’t publish over latest (if we want to do that)?

gksander:
Yeah, I was a bit curious on the tags as well... I manually bumped the version instead of using changesets, figured we might just publish manually until we have room to explore how changesets will interact with the tag.

carlos-kelly (changes requested):
Only doing a block so we don't miss adding `private: true` to the package.json for the CLI to prevent a publish on this one yet.

carlos-kelly (approved):

PR #1180, opened 2022-08-02
removing unused dependencies

### LINKED ISSUE #1014: Chore: Audit prod dependencies
We should audit all our prod deps for actual usage. Discovered today that we're not using:

- [ ] `marksy`: https://github.com/FormidableLabs/spectacle/search?q=marksy

paulmarsicloud:
### Description

Removing

```
broadcastchannel-polyfill
dedent
prism-react-renderer
```

As these don't appear to be used per `npx depcheck` and `npx npm-check`

Updated `CONTRIBUTING.md` for minor fix

Fixes #1014 

#### Type of Change

...although this is more like a chore than a bug 🤷 haha

### How Has This Been Tested?

Ran through all the tests listed [here](https://github.com/FormidableLabs/spectacle/blob/main/CONTRIBUTING.md#examples) and everything built so this should be safe 👍


packages/spectacle/package.json
```diff
@@ -37,15 +37,12 @@
   "dependencies": {
     "@types/testing-library__jest-dom": "^5.14.5",
     "broadcast-channel": "^3.2.0",
-    "broadcastchannel-polyfill": "^1.0.0",
```
gksander, line 40: Nice, looks like we now have this:

```ts
import { BroadcastChannel as BroadcastChannelPolyfill } from 'broadcast-channel';

// ...

const BroadcastChannel = window.BroadcastChannel || BroadcastChannelPolyfill;
```

and no longer using the `broadcastchannel-polyfill` lib

packages/spectacle/package.json
```diff
@@ -37,15 +37,12 @@
...
-    "dedent": "^0.7.0",
     "history": "^4.9.0",
     "kbar": "0.1.0-beta.36",
     "mdast-builder": "^1.1.1",
     "mdast-zone": "^4.0.0",
     "merge-anything": "^3.0.3",
     "mousetrap": "^1.6.5",
-    "prism-react-renderer": "^0.1.7",
```
gksander, line 48: Nice, looks like we're using `react-syntax-highlighter` instead of this.

packages/spectacle/package.json
```diff
@@ -37,15 +37,12 @@
   "dependencies": {
     "@types/testing-library__jest-dom": "^5.14.5",
     "broadcast-channel": "^3.2.0",
-    "broadcastchannel-polyfill": "^1.0.0",
-    "dedent": "^0.7.0",
```
gksander, line 41: Nice, looks like we have our own `indentNormalizer` method that I'm assuming replaced this but we never got rid of the dep.

gksander (approved):
Nice finds! Thanks, Paul!

PR #1185, opened 2022-08-04
Infra: build create-spectacle examples

ryan-roemer:
Core work:
- Add create/install/build/start helpers for create-spectacle
- Add CI support

Work along the way:

- Extract `jest` dependencies to monorepo root
- Lots of wireit configuration cleanup


.eslintrc
```diff
@@ -36,13 +36,6 @@
     }
   },
   "rules": {
-    "quotes": [
```
ryan-roemer, line 39: This should be handled by prettier and prettier _sometimes_ wants double quotes so disabling here.

package.json
```diff
@@ -23,16 +26,29 @@
...
+    "ts-jest": "^28.0.7",
     "typescript": "^4.7.4",
+    "wait-on": "^6.0.1",
     "webpack": "^5.68.0",
     "webpack-cli": "^4.10.0",
     "webpack-dev-server": "^4.7.4",
     "wireit": "^0.7.1"
   },
+  "pnpm": {
+    "neverBuiltDependencies": [
```
ryan-roemer, line 48: This prevents puppeteer from trying run husky (which isn't even a prod dep) on installation. We then separately have chromium installation instructions.

packages/create-spectacle/package.json
```diff
@@ -35,7 +35,24 @@
     "lint": "wireit",
     "lint:fix": "wireit",
     "prettier": "wireit",
-    "prettier:fix": "wireit"
+    "prettier:fix": "wireit",
+    "examples:clean": "rimraf .examples",
+    "examples:test": "nps jest",
+    "examples:jsx:clean": "rimraf .examples/jsx",
```
ryan-roemer, line 41: I'm not totally in love with this verbose enumeration of tasks. But it does make scripting in CI much simpler... Open to other suggestions!

packages/create-spectacle/package.json
```diff
@@ -35,7 +35,24 @@
     "lint": "wireit",
     "lint:fix": "wireit",
     "prettier": "wireit",
-    "prettier:fix": "wireit"
+    "prettier:fix": "wireit",
+    "examples:clean": "rimraf .examples",
+    "examples:test": "nps jest",
```
ryan-roemer, line 40: There's just one test that assumes a final "made with" slide from a dev server running on port 3000. Open to other more flexible suggestions!

packages/create-spectacle/tsconfig.json
```diff
@@ -4,6 +4,8 @@
     "moduleResolution": "node",
     "module": "commonjs",
     "outDir": "./bin",
+    // https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping#node-14
+    "target": "ES2020",
```
ryan-roemer, line 8: @carlos-kelly I've updated the spectacle bin output to be modern supported node target (14)

gksander (approved):
Looks great! I wonder if it makes sense to only run this on changes to `packages/create-spectacle` 🤔 I guess we can monitor and see how long this takes to run, and if it ends up being any sort of bottleneck for general PRs, we could split it out then.

packages/create-spectacle/package.json
carlos-kelly: we could port forward but that feels over engineered here. that to me is just testing if we use `commander` correctly rather than the output correctly building and not having runtime errors.

carlos-kelly (approved):
This is really great.

PR #1186, opened 2022-08-09
[FIX] Issue-1171 Do not overwrite url derived state

### LINKED ISSUE #1171: URLs for slide broken, slideshow always starts at 0 on load/refresh.
### Describe Your Environment

What version of Spectacle are you using? (can be found by running `npm list spectacle`)

Fresh clone & install of https://github.com/FormidableLabs/spectacle/commit/8dac6a30fac55e4c65fe23288ee80964b4e536f6

(by the way this is broken too. had to manually install pnpm)

```bash
yarn add pnpm
npx pnpm install
npx pnpm start:js
```

What browser are you using?
Tested on Firefox and Chrome

What machine are you on?
macOS and android

### Describe the Problem

True of any slideshow I try, including the examples in the repository:

Opening a slideshow and navigating to another slide, eg. `http://localhost:3000/?slideIndex=2&stepIndex=0`, and then refreshing will take me back to the 0th slide. Going to the link directly also boots me back to the beginning of the slideshow. This sort of sucks a lot and breaks one of the major reasons why I use this -- being able to link directly to specific slides.

I think it might have something to do with slide 0, step 0 being hardcoded in the deck's usage of location sync....

https://github.com/FormidableLabs/spectacle/blob/8dac6a30fac55e4c65fe23288ee80964b4e536f6/packages/spectacle/src/components/deck/deck.tsx#L281-L287

edit: was able to resolve the problem by reverting to the old style ReactDOM usage. 

ie. rather than

```jsx
import { createRoot } from 'react-dom/client';

const container = document.getElementById('root')
const root = createRoot(container);
root.render(<Presentation />);
```

using 

```jsx
import ReactDOM from 'react-dom'

ReactDOM.render(<Presentation />, document.getElementById('root'))
```

AaronPowell96:
### Description

I have removed an effect that seemed to be overwriting the state derived from the url. 
I understand that Spectacle's approach is to maintain its own state whereby location is superseded by app state, however, I believe in this instance on initialisation we do want location to "win".

I hope that someone with a bit more context on this effect will have input on the impacts of its removal.

As far as I can tell, we can still set initial state once upon mount by spreading the argument passed to useDeckState. The only logical part that is no longer checked is ``` if(initialized) ``` I do think at mount (when passing state to the reducer) we are not initialised so this is fine. However, something to think about for someone who has context.

Fixes #1171 

#### Type of Change

- [ X] Bug fix (non-breaking change which fixes an issue)

### How Has This Been Tested?

Ran all current tests and manually ran each example to see if they continued to work.

<details>
<summary>Before</summary>

https://user-images.githubusercontent.com/17838632/183662555-696d1b75-463b-4103-b1b9-ba464d522599.mov

</details>

<details>
<summary>After</summary>

https://user-images.githubusercontent.com/17838632/183662734-04a5be15-f29c-4eab-8590-bcaba21b5082.mov

</details>

### DIFF
--- a/packages/spectacle/src/hooks/use-deck-state.ts
+++ b/packages/spectacle/src/hooks/use-deck-state.ts
@@ -100,7 +100,7 @@ function deckReducer(state: DeckState, { type, payload = {} }: ReducerActions) {
 export default function useDeckState(userProvidedInitialState: DeckView) {
   const [{ initialized, pendingView, activeView }, dispatch] = useReducer(
     deckReducer,
-    initialDeckState
+    { ...initialDeckState, ...userProvidedInitialState }
   );
   const actions = useMemo(
     () => ({
@@ -119,12 +119,6 @@ export default function useDeckState(userProvidedInitialState: DeckView) {
     }),
     [dispatch]
   );
-
-  useEffect(() => {
-    if (initialized) return;
-    if (userProvidedInitialState === undefined) return;
-    actions.initializeTo(userProvidedInitialState);
-  }, [initialized, actions, userProvidedInitialState]);
 
   return {
     initialized,


carlos-kelly (approved):
Thanks for this bug fix. I'll make a follow up to handle out-of-bounds URL protection.

PR #1192, opened 2022-08-10
`create-spectacle` Spectacle version fix.

gksander:
## What's the issue?

Running `yarn create spectacle` results in this error:

```
Error: Cannot find module 'spectacle/package.json'
Require stack:
- /Users/gksander/.config/yarn/global/node_modules/create-spectacle/bin/cli.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
```

This is due to us trying to import `package.json` from `spectacle`, which is not installed.


packages/create-spectacle/package.json
```diff
@@ -22,7 +22,7 @@
   },
   "peerDependencies": {},
   "devDependencies": {
-    "spectacle": "*",
+    "spectacle": "workspace:*",
```
gksander, line 25: Using `workspace:*` tells `pnpm` to replace `workspace:*` with actual `spectacle` version during packaging, whereas the `"spectacle": "*"` is preserved as-is during packaging.

ryan-roemer: Put the _actual version_ in and changesets will update the version if specified in `linked` in changsets config.

packages/create-spectacle/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "create-spectacle",
   "version": "0.1.0",
   "description": "Project generator for Spectacle",
-  "main": "bin/cli.js",
+  "main": "bin/src/cli.js",
```
ryan-roemer, line 5: This should be `bin/cli.js`. See https://unpkg.com/browse/create-spectacle@0.1.0/bin/

packages/create-spectacle/package.json
gksander: My understanding is, if we specify a linkage between the two, then any time we do a minor version bump of `create-spectacle`, we'll also be making a minor version bump of `spectacle` (and so on). Seems like we might not want that right away.

packages/create-spectacle/src/cli.ts
```diff
@@ -9,7 +9,7 @@ import {
   writeWebpackProjectFiles,
   writeOnePageHTMLFile
 } from './templates/file-writers';
-import { version as SPECTACLE_VERSION } from 'spectacle/package.json';
+import { version as SPECTACLE_VERSION } from '../package.json';
```
ryan-roemer, line 12: We don't want the `create-spectacle` version, we want something like:

```suggestion
import { devDependencies: { spectacle as SPECTACLE_VERSION } } from '../package.json';
```

packages/create-spectacle/package.json
gksander: > Since the `package.json` file is used, the new TS root dir becomes `.` instead of `src`, so the bin location is now nested.

packages/create-spectacle/src/cli.ts
ryan-roemer, line 12: We don't want the `create-spectacle` version, we want something like:

```suggestion
import { devDependencies: { spectacle as SPECTACLE_VERSION } } from '../package.json';
```

packages/create-spectacle/package.json
ryan-roemer: But as long as it works (maybe we run a quick publish experiment in the experiments repo?

ryan-roemer (approved):
OK, so I _think_ this all should work (e.g. that changesets will correctly replace `workspace:*` with version on release)

I'm not totally sure, however, if we want to do `linked` for `spectacle` and `create-spectacle` per https://github.com/changesets/changesets/blob/main/docs/linked-packages.md

PR #1194, opened 2022-08-10
Infra: Separate out examples CI steps

ryan-roemer:
@gksander and I noticed that our examples create/install/build/start single run step was weirdly failing midway through without actually failing the action, leaving the test action to hang for a long time before failing. This PR separates out those steps so that we have isolated insight into what's happening and if the install fails silently, the next build step should fail the workflow much faster than before.


.github/workflows/create-spectacle.yml
```diff
@@ -5,11 +5,13 @@ on:
     branches:
       - main
     paths:
+      - ".github/workflows/create-spectacle.yml"
```
ryan-roemer, line 8: We should have had this anyways

gksander: Nice!

gksander (approved):
🚀

ryan-roemer:
Idea -- I think what might be happening is `wait-for` is getting _some_ response from the webpack dev server, but not sufficient for Jest testing. Maybe switch to static serving the built content or putting a sleep after starting the dev server before wait-for?

gksander (approved):
This looks good to me.

PR #1195, opened 2022-08-11
`create-spectacle` prompts

gksander:
## `create-spectacle` prompts

This PR adds `prompts` to the `create-spectacle` package, so that users do not need to remember flags (although flags are still available). Some things to note:

- If you pass all possible flags, and you're not trying to overwrite a folder, we bypass prompts.
- All four prompts (name, type, lang, port) are shown – but have default values based on args/flags. 
- I tried damn hard to test this, but woah. What a challenge. Writing tests around dynamic stdout/stdin with these interactive prompts was tough and flakey and I just couldn't pull it off in the time I allotted for it. Something I wouldn't mind looping back to, but decided to cut my losses for the week.

### Prompt flow 

cs-prompt1

### All flags, no prompts

cs-prompt2


carlos-kelly (commented):
In `file-writers.ts` let's remove `await rm(outPath, { recursive: true, force: true });` which deletes the directory no matter what.

carlos-kelly (commented):
This looks great but let's remove `mdx` for now since it doesn't generate anything yet. That's the last part to implement for both flags and prompts.

packages/create-spectacle/src/cli.ts
```diff
@@ -12,49 +16,126 @@ import {
...
+      const response = await prompts(
+        [
+          // Name prompt
+          {
+            type: 'text',
+            name: ArgName.name as string,
+            message: 'What is the name of the presentation?',
+            initial: name,
+            validate: async (val) => {
+              return val.length > 0 ? true : 'Name is required';
```
carlos-kelly, line 71: worth running `.trim()` too?

packages/create-spectacle/src/cli.ts
```diff
@@ -12,49 +16,126 @@ import {
...
+          },
+          // Language prompt
+          {
+            type: 'text',
+            name: ArgName.lang as string,
+            message:
+              'What is the language code for the generated HTML document?',
+            initial: lang,
+            validate: async (val) => {
+              return val.length > 0 ? true : 'Language code is required';
```
carlos-kelly, line 111: Should we require it? Or make `en` a default like the flags? Typically I've seen [en] default and pressing enter will go with that

packages/create-spectacle/src/cli.ts
gksander: Good idea!

packages/create-spectacle/src/cli.ts
gksander: So the `initial: lang` bit will default it, so you can just hit Enter to select `en` by default. This validation is if the user Backspaces twice to try to use `""` as the input.

packages/create-spectacle/README.md
```diff
@@ -1,3 +1,29 @@
...
+```
+
+Once running the respective command, you will be prompted to provide information about the spectacle project you'd like to create. Once you provide necessary information, a new spectacle project will be created in the directory derived from the project name you provided.
+
+## Flags
+
+`create-spectacle`'s core usage is via the interactive prompts. However, there are a handful of arguments/flags that you can provide to pre-fill prompt options:
+
+- Pass a project name as the main argument to specify a project name, e.g. `yarn create spectacle my-presentation`.
+- Pass the `--type` or `-t` flag to specify the type of spectacle project you'd like to create. Options are `jsx`, `tsx`, `mdx`, or `onepage`. Example: `yarn create spectacle -t onepage my-presentation`.
```
carlos-kelly, line 19: Can we drop the mdx ref here too?

gksander: Can do!

gksander: Consider it done!

carlos-kelly (approved):

Issue #1201, opened 2022-08-12
`website`: Typechecking doesn't work.


ryan-roemer:
Reproduction

```sh
$ cd website
$ yarn install
$ yarn typecheck
```

paulmarsicloud:
I think we're good to close this once https://github.com/FormidableLabs/spectacle/pull/1216 is merged right?

PR #1202, opened 2022-08-12
Infra: Add docs to pnpm

ryan-roemer:
Work:

- Add `website` to part of pnpm workspaces.
- Move MJS files to CJS
- Add some missing dependencies

Work for later:

- [ ] Add `lint` and `prettier` and make part of main CI task with wireit, etc. I'm not doing it now because it may touch a lot of files and confuse this PR. https://github.com/FormidableLabs/spectacle/issues/1203
- [ ] Add typechecking to main CI as well -- this isn't working at all currently -- see #1201


website/package.json
```diff
@@ -19,13 +17,16 @@
     "write-heading-ids": "docusaurus write-heading-ids",
     "typecheck": "tsc"
   },
-  "dependencies": {
+  "devDependencies": {
```
ryan-roemer, line 20: We have no actual runtime -- everything's a devDep. We could consider the App things to be "prod deps" if we want and I can refactor, but the current setup in main has no consistency...

jpdriver (approved):
i guess this is fine to come in, but i _did_ want to some day share the `docs` workflow..

maybe one for later!

gksander (approved):
👍

Issue #1199, opened 2022-08-12
`create-spectacle`: Overwriting for onepage vs t/jsx.


gksander:
A `onepage` project ends up in a file `foobar.html`, whereas the `jsx/tsx` project ends up in `foobar/`.

If user provides a project name `foobar` and `foobar/` directory already exists, we prompt them regarding overwriting the directory. However, if the `foobar/` directory already exists, but the user selects a `onepage` project, then we don't necessarily need to overwrite anything.

We should make the overwrite option more flexible based on the type of project. We likely need to prompt the user for type of project prior to asking for the name, so we can do a dynamic overwrite check based on type of project.

gksander:
Closed via #1205

PR #1205, opened 2022-08-15
`create-spectacle`: Better check for overwrite

gksander:
Address #1199 

Since `onepage` generates a single HTML file, but other deck types create a directory – we need to have a bit of a logic fork for determining whether or not we'll be overwriting (and when to prompt the user to confirm an overwrite). The approach here is to just move the `type` prompt right after `name` (but before the overwrite confirmation), so we can appropriately determine _when_ we need to overwrite (based on both `type` and `name`).

### DIFF
--- a/packages/create-spectacle/src/cli.ts
+++ b/packages/create-spectacle/src/cli.ts
@@ -46,8 +46,8 @@ const main = async () => {
   let i = 0;
   let type = argv[ArgName.type] || argv['t'];
   let name = argv['_']?.[0];
-  let lang = argv[ArgName.lang] || argv['l'] || 'en';
-  let port = argv[ArgName.port] || argv['p'] || 3000;
+  let lang = argv[ArgName.lang] || argv['l'];
+  let port = argv[ArgName.port] || argv['p'];
 
   const isTryingToOverwrite =
     Boolean(name) && !isOutputPathAvailable(name, type === 'onepage');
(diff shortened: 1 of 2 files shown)


packages/create-spectacle/src/cli.ts
```diff
@@ -46,10 +46,11 @@ const main = async () => {
   let i = 0;
   let type = argv[ArgName.type] || argv['t'];
   let name = argv['_']?.[0];
-  let lang = argv[ArgName.lang] || argv['l'] || 'en';
-  let port = argv[ArgName.port] || argv['p'] || 3000;
+  let lang = argv[ArgName.lang] || argv['l'];
+  let port = argv[ArgName.port] || argv['p'];
```
gksander, line 50: Moving these fallbacks to the prompts. Result of this...

* **before this change** user could pass `type` and `name`, and they wouldn't be prompted at all – just straight to creating decks.
* **after this change** user can pass `type` and `name`, but they'll still get prompted to confirm lang/port. If _all_ of the values are explicitly provided via flags (and no overwrite detected), prompts will be bypassed.

gksander:
@carlos-kelly just a friendly little ping. I think after this, most of the foundational work is in place for `create-spectacle`.

packages/create-spectacle/src/cli.ts
```diff
@@ -72,14 +74,43 @@ const main = async () => {
...
+              const name = answers?.[ArgName.name];
+              const type = answers?.[ArgName.type];
+              if (type === 'onepage') {
+                return `File ${formatProjectOutputPath(
+                  name
+                )}.html already exists. Overwrite it?`;
+              } else {
+                return `Target directory ${formatProjectOutputPath(
+                  name
+                )} already exists. Overwrite and continue?`;
```
carlos-kelly, line 111: This is really nit picky but we have two different prompt strings for overwriting. Let's standardize on "Override and continue?" for line 107 as well.

carlos-kelly (approved):
This is great. Just one suggestion on user-facing language then merge away!

packages/create-spectacle/src/cli.ts
gksander: Ooooh, detailed eye! Nice!

PR #1206, opened 2022-08-16
Add `types:create:watch` to example scripts

gksander:
The `start:ts` etc scripts are great for development, as it allows you to develop library code alongside example code for a fantastic feedback cycle.

However, currently, these scripts are only building the ES files, and we're missing updated type definitions. This PR adds `pnpm:build:spectacle:types:create:watch` to these `start:` scripts so we get updated type definitions as we develop.


ryan-roemer (changes requested):
Nice bug find!

For this PR, we don't have `build:spectacle:types:create:watch` in `package.json:scripts` (like, e.g. `build:spectacle:esm:watch`) and need to add it. The error message of:

```
 ERR_PNPM_NO_SCRIPT  Missing script: build:spectacle:types:create:watch
```

gets wiped out in `concurrently --raw` mode because the webpack dev server rewrites the terminal.

ryan-roemer (approved):
LGTM!

PR #1210, opened 2022-08-19
`create-spectacle`: vite starters, add simple-ish test regarding file generation

gksander:
This PR adds `vite` starters for `create-spectacle`, since `vite` is speedy and easy and cool, and TBH my preferred build tool if I'm cooking up a Spectacle project.

Also added tests around file generation. I didn't intended for these to be as full-fledged as our actual e2e tests here, just wanted to make sure we're generating appropriate files and they have some key content that we expect.


Manubi:
Can we add this? :D

gksander:
@carlos-kelly curious about your thoughts on this! Kind of a beefy PR, but most of the LoC is in the new tests.

carlos-kelly (commented):
Let's add some npm scripts that mirror the other output types in this package's package.json

carlos-kelly:
This is great and works well. We can adjust the prompt colors in a follow up PR.

carlos-kelly (approved):

PR #1209, opened 2022-08-18
feat: Add new slide layouts: section, statement, big fact, quote.

ZimChristine:
Description

The changes in the PR add four new types of Slide Layouts: Section, Statement, Big fact, and Quote. These layouts were modeled after Keynote slide templates with the goal of making slide creation quicker and easier for common use cases.

Type of Change
[ x ] New feature (non-breaking change which adds functionality)
How Has This Been Tested?
Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.
The changes include adding new tests in packages/spectacle/src/components/slide-layout.test.tsx
While developing the feature, I manually consumed the layout components in the Presentation deck in index.tsx

Checklist: (Feel free to delete this section upon completion)
[ n/a ] I have included a [changeset](https://github.com/FormidableLabs/spectacle/CONTRIBUTING.md#changesets) if this change will require a version change to one of the packages.
[ x ] I have performed a self-review of my own code
[ x ] I have commented my code, particularly in hard-to-understand areas
[ x ] I have made corresponding changes to the documentation
[ x ] I have run pnpm run check:ci and all checks pass
[ x ] I have added tests that prove my fix is effective or that my feature works
[ x ] My changes generate no new warnings
[ n/a ] Any dependent changes have been merged and published in downstream modules

PR #1213, opened 2022-08-23
Slide layout tweaks

gksander:
This PR builds off of #1209, tweaks just a couple of very minor things:

- For `SlideLayout.BigFact`, use `children` for the fact prop (instead of a separate prop).
- For `SlideLayout.Quote`, use `children` for the quote prop (instead of a separate prop).
- Updates docs accordingly.


carlos-kelly (approved):
Love the migration to using `children`. I'm not a huge fan of elements with a single render prop.

PR #1217, opened 2022-08-31
feat: Add single and multiple code pane layouts

ZimChristine:
### Description

The changes in this PR add two new slide layouts for easily adding a code with syntax highlighting and a title. There are two slide layouts: 
1) a single code pane with optional title
2) multiple code panes with optional descriptions and an optional title

#### Type of Change

- [ x ] New feature (non-breaking change which adds functionality)

### How Has This Been Tested?

* Jest tests have been added for the feature.
* During development the feature was manually tested by consuming the new `SlideLayouts` in the example `tsx` app.

### Checklist: 

- [ x ] I have included a [changeset](../CONTRIBUTING.md#changesets) if this change will require a version change to one of the packages.
- [ x ] I have performed a self-review of my own code
- [ x ] I have commented my code, particularly in hard-to-understand areas
- [ x ] I have made corresponding changes to the api documentation
- [ x ] I have run `pnpm run check:ci` and all checks pass
- [ x ] I have added tests that prove my fix is effective or that my feature works
- [ x ] My changes generate no new warnings
- [ na ] Any dependent changes have been merged and published in downstream modules


docs/api-reference.md
```diff
@@ -447,4 +447,28 @@ A vertically-centered Quote layout for if you want to present a quote and attrib
...
+### `SlideLayout.MultiCodeLayout`
+
+A layout with multiple code panes and optional descriptions, and an optional title for if you want more than one code block per slide or code with description text.
+
+| Props              | Type                                | Required | Example                         |
+|--------------------|-------------------------------------|----------|---------------------------------|
+| `...slideProps`    | [Slide Props](#slide)               | ❌        |                                 |
+| `title`            | `string`                            | ❌        | `Show me the code!`             |
+| `titleProps`       | [Heading Props](#typography-tags)   | ❌        | `{ color: 'red' }`              |
+| `children`         | `PropTypes.arrayOf(PropTypes.object([Code Pane props](#code-pane)))` | ✅        | `[{ code: 'console.log("hello world!")', language: 'jsx', description: 'Say hello', props: {...} }, {...}]` |
```
ZimChristine, line 473: I'm unsure if the current Type column for `codePaneProps` is the best way to document the type for this prop.

docs/api-reference.md
```diff
@@ -447,4 +447,27 @@ A vertically-centered Quote layout for if you want to present a quote and attrib
...
+### `SlideLayout.MultiCodeLayout`
+
+A layout with multiple code panes and optional descriptions, and an optional title for if you want more than one code block per slide or code with description text.
+
+| Props              | Type                                | Required | Example                         |
+|--------------------|-------------------------------------|----------|---------------------------------|
+| `...slideProps`    | [Slide Props](#slide)               | ❌        |                                 |
+| `title`            | `string`                            | ❌        | `Show me the code!`             |
+| `titleProps`       | [Heading Props](#typography-tags)   | ❌        | `{ color: 'red' }`              |
+| `codePaneProps`    | `PropTypes.arrayOf(PropTypes.object([Code Pane props](#code-pane)))` | ✅        | `[{ code: 'console.log("hello world!")', language: 'jsx', description: 'Say hello', props: {...} }, {...}]` |
```
ZimChristine, line 473: I'm unsure if the current Type column for `codePaneProps` is the best way to document the type for this prop.

packages/spectacle/src/components/slide-layout.tsx
```diff
@@ -195,11 +196,117 @@ const Quote = ({
...
+
+/**
+ * multiple Code Panes with optional Description, with optional Title layout
+ */
+const MultiCodeLayout = ({
+  codePaneProps,
+  title,
+  titleProps,
+  ...rest
+}: Omit<SlideProps, 'children'> & {
```
gksander, line 268: I like this, but have a couple thoughts. 

First, I think `codePanProps` might lead to confusion – since they aren't really props for `CodePane` items. Maybe we can shorten that to something like `codeBlocks` or something? (Idk, naming is hard.)

Next, what do you think about offering a grid layout so a user could choose between side-by-side and vertically stacked?

image

If you think that makes sense, here's what I used for that:

```tsx
const MultiCodeLayout = ({
  codePaneProps,
  title,
  titleProps,
  numColumns = 1,
  ...rest
}: Omit<SlideProps, 'children'> & {
  codePaneProps: {
    code: string;
    language: string;
    description?: string | ReactNode;
    descriptionProps?: ComponentProps<typeof Text>;
  }[];
  title?: string | ReactNode;
  titleProps?: ComponentProps<typeof Text>;
  numColumns?: number;
}) => {
  return (
    <Slide {...rest}>
      {title ? <Heading {...titleProps}>{title}</Heading> : null}
      <Grid
        gridRowGap={1}
        gridColumnGap={1}
        gridTemplateColumns={`repeat(${numColumns}, 1fr)`}
        maxWidth="100%"
      >
        {codePaneProps.map(
          ({ code, language, description, descriptionProps, ...props }, i) => (
            <CodeLayout
              key={i}
              text={description}
              textProps={descriptionProps}
              language={language}
              {...props}
            >
              {code}
            </CodeLayout>
          )
        )}
      </Grid>
    </Slide>
  );
};
```

I had to also delete [this line](https://github.com/FormidableLabs/spectacle/blob/605436120df0d5694d9a1210b9ea0d0675d5bc33/packages/spectacle/src/components/code-pane.tsx#L160) to get that working, but I think we should maybe have a fluid width for code panes anyways 🤔 

Happy to chat more about this! Let me know what you think!

gksander (changes requested):
Looking great, I can't wait to use this! A couple little suggestions here.

packages/spectacle/src/components/slide-layout.tsx
ZimChristine: Glad you like it! I think option to use a grid layout is a good idea. I'll make the changes.

ZimChristine: > First, I think codePanProps might lead to confusion – since they aren't really props for CodePane items. 

These properties do get passed to the `CodePane` component which is why I thought naming them as such would reduce confusion. Leaving the name the same for now, but happy to continue the discussion.

packages/spectacle/src/components/slide-layout.tsx
```diff
@@ -195,11 +196,117 @@ const Quote = ({
...
+}) => {
+  return (
+    <Slide {...rest}>
+      <FlexBox
+        style={{
+          display: 'flex',
+          flexDirection: 'column',
+          height: '100%'
+        }}
+      >
```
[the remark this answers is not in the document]
ZimChristine: I debated on whether to default to top or center justified and ended up following what seemed to me the more prevalent pattern of center. Keynote mostly does center-justified and that's what the other new Spectacle Slide Layouts use too.

🌞 I don't have a strong opinion about this, so let's do top-justified!

packages/spectacle/src/components/slide-layout.tsx
```diff
@@ -195,11 +196,117 @@ const Quote = ({
...
+        {title ? <Heading {...titleProps}>{title}</Heading> : null}
+        <CodeLayout language={language} {...rest}>
```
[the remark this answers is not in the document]
ZimChristine: Oh. My goal was to allow other CodePane component props through, e.g. `highlightRange`. Help me understand how I ought to have accomplished this?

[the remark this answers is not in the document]
gksander: In `code-pane.tsx` we have an exported type `CodePaneProps` for the props of the `CodePane` components. To pass props along to the `CodePane` element, I'd recommend having another prop on this layout specifically for these props. This is where I think a `codePaneProps` would make sense.

packages/spectacle/src/components/slide-layout.tsx
gksander: It looks like some of these props get passed through `CodeLayout` into `CodePane` – but not _all_ of these fields get passed all the way through to `CodePane`, and there are some `CodePane` props that don't exist here.

For example, having `descriptionProps` inside of `codePaneProps` feels a bit confusing to me, since we have an explicit prop type:

```ts
export type CodePaneProps = {
  children: string;
  language: string | undefined;
  theme?: Record<string, CSSProperties>;
  stepIndex?: number;
  highlightRanges?: Ranges;
  showLineNumbers?: boolean;
};
```

packages/spectacle/src/components/slide-layout.tsx
```diff
@@ -195,11 +196,124 @@ const Quote = ({
...
+        {title ? <Heading {...titleProps}>{title}</Heading> : null}
+        <CodeLayout language={language} {...codePaneProps}>
+          {children}
+        </CodeLayout>
+      </Box>
+    </Slide>
+  );
+};
+
+interface CodeBlocks extends CodePaneProps {
```
ZimChristine, line 256: Here's what I came up with to deal with `CodePaneProps` from our [previous discussion about CodePaneProps](https://github.com/FormidableLabs/spectacle/pull/1217#discussion_r961613775).

I would have liked to extend the CodePaneProps (inline down here)[https://github.com/FormidableLabs/spectacle/pull/1217/files#diff-c1ac090b2ff636b65120246c862260a07e29ceacc75778bc3c88ec0d89868a9bR270] but how to do that gracefully was lost on me. @gksander, do you see a more elegant way of doing this? Thanks!

packages/spectacle/src/components/code-pane.tsx
```diff
@@ -149,15 +149,13 @@ const CodePane = forwardRef<HTMLDivElement, CodePaneProps>(
        * default theme with no valid values.
        */
       const {
-        size: { width = 1366 },
```
ZimChristine, line 152: 🎗 Reminder to self to test this change before merging this PR.

packages/spectacle/src/components/slide-layout.tsx
ZimChristine: Yes! You're exactly right and I agree it was confusing. I fixed this to use the exported Code Pane Props type and extended it for `CodeBlocks`

packages/spectacle/src/components/code-pane.tsx
ZimChristine: I am uncertain what controlling `width` here is trying to achieve (or prevent) so I'm hesitant to want to make this change before knowing what I'm deleting. Sending an empty theme object through (e.g. `theme={{}}`) doesn't appear to affect much change to the codepane.

Maybe you can help me understand the context more fully @gksander ? THANKS!

gksander: This looks good, as per our discussion.

gksander (approved):
🎈 LGTM

PR #1219, opened 2022-09-10
Fix Markdown rendering inside templates.

carlos-kelly:
### Description

Fixes a bug where a Markdown component inside a Deck template would break Spectacle. The previous implementation of Markdown always creates a Notes node tree, even with undefined children if there are no notes, which cannot exist on a template as it's not a slide. This verifies there is note content before building the Notes node tree.

#### Type of Change

### How Has This Been Tested?

Add a `<Markdown />` component with text content to a Deck template before this change and see Spectacle break and fail to render. Check it after and see it correctly render the Markdown component. 

Example below showing a Markdown component on the template and notes inside a Markdown slide verifying no regressions introduced.

Screen Shot 2022-09-10 at 4 10 19 PM

### DIFF
--- a/.changeset/tiny-buses-fold.md
+++ b/.changeset/tiny-buses-fold.md
@@ -1,0 +1,5 @@
+---
+'spectacle': patch
+---
+
+Fixed Notes node tree generation inside Markdown component.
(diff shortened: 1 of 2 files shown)


gksander (approved):
LGTM

PR #1222, opened 2022-09-15
CI tweaks investigation

fritz-c:
Investigating solutions to the issues in #1194


fritz-c:
Moving these commits to the original PR.

PR #1223, opened 2022-09-15
Clean up issue and PR templates

fritz-c:
### Description

I converted most of the instructional text for each section into comments, because the ambiguity about whether people should delete them or not usually results in a lot of them ending up in issues that are submitted, adding to visual clutter and making the important information harder to find.

I also updated the codesandbox links that are referenced to a couple that have been updated to the latest dependencies, one for JS and one for TS. Those are currently under my personal codesandbox account, but we can fork them over to any other (company?) account if there are concerns about maintaining them. 

#### Type of Change

Github-facing only

### How Has This Been Tested?

N/A


gksander (approved):
Nice, this is good cleanup!

PR #1231, opened 2022-09-21
Fix: theme.backdropStyle overrides

### LINKED ISSUE #1051: What is the correct way to replace the backdrop color?
I'd like to have a white background in my slides, and I'd prefer a fadeTransition, though slideTransition has the same problem. Best way to put it: the transitions go through a black phase, no matter what I tried.

In the source code I found ways to specify backdropStyle in the theme settings. That seems to work in presenter mode, but not in the normal mode.
```javascript
const theme = {
  fonts: {
    header: '"Ubuntu", Helvetica, Arial, sans-serif',
    text: '"Fira sans", Helvetica, Arial, sans-serif'
  },
    colors: {
      primary: '#000000',
      secondary: '#0000ff',
      tertiary: '#ffffff',
      background:"#ffffff"
    },
  backdropStyle:{
    backgroundColor:"#ffffff",
    position: 'fixed',
    top: 0,
    left: 0,
    width: '100vw',
    height: '100vh'
}
}
```

This is also not covered in the documentation. Any help would be appreciated.

fritz-c:
### Description

There is currently no straightforward way to override the background color of the presentation backdrop.

* The code seems to have been designed to accommodate custom background colors, but it was missing logic that would prevent values set in `theme.backdropStyle` from being overridden by the default black color.
* The prop that had implementation allowing it to override the default background color, `backdropStyle`, is accessible from the `Deck` props, but not present in the TypeScript types, so [TypeScript users had to use a workaround](https://github.com/FormidableLabs/spectacle/issues/1051#issuecomment-1248540386) to push it through.

My judgement was that there was no need to have multiple ways to accomplish the same outcome with the styling props, as it complicates the implementation and the API. I ultimately kept the current behavior the same, except for the bug that causes `theme.backdropStyle.background[Color]` to be ignored, but marked the excess props (`backdropStyle`, `suppressBackdropFallback`, `theme.suppressBackdropFallback`) as deprecated in the TypeScript types, to be removed in the next major release.

Fixes #1051

#### Type of Change

### How Has This Been Tested?

New unit tests


packages/spectacle/src/components/deck/deck.tsx
```diff
@@ -400,21 +400,16 @@ export const DeckInternal = forwardRef<DeckRef, DeckInternalProps>(
     //
     // Yes, this is slightly awkward, but IMO adding an additional `<div>` element
     // would be even more awkward.
-    let useFallbackBackdropStyle = true;
-    const backdropStyle = themeProvidedBackdropStyle;
-    let BackdropComponent = 'div' as ElementType;
-    if (userProvidedBackdropStyle) {
-      Object.assign(backdropStyle, userProvidedBackdropStyle);
```
fritz-c, line 407: Probably never happened to a real user in a way that would surface as a bug, but this `Object.assign` call does an in-place modification of the object the user passes into `theme.backdropStyle`.

carlos-kelly (approved):
This is great. I wonder if the documentation should clearly define what a Backdrop is, aka, the thing rendered behind the slides. And sometimes if your browser is sized at the same ratio of your slides you might not even see it.

Issue #1260, opened 2023-05-03
Remove `create-react-app` option, Add new option 1 using Next.js?


karlhorky:
### Description

Now that [`create-react-app` is going away](https://github.com/reactjs/react.dev/pull/5487#issuecomment-1409720741) (it's turning into a launcher), I would propose the following changes:

1. Remove all mentions of `create-react-app` from the tutorial (this is Option 1, so very visible)
2. Add a new Option 1 that uses Next.js instead - ideally using the App Router + React Server Components

### Proposal

By changing the documentation and providing a polished, tested template using Next.js

### Links / References

Linked above already

carloskelly13:
@karlhorky This is a good idea. I think the quickest way to boot strap a new Spectacle deck when wanting to use JSX + React components is Vite, but Next App Router is also a good choice.

karlhorky:
Thanks for this @carloskelly13 ! 🙌 @ProchaLu will try out migrating our slides from `mdx-deck` to Spectacle + Next.js + App Router.

carloskelly13:
Yeah! Let us know if you find anything during the migration!

PR #1263, opened 2023-05-06
Added support for App Router + SSR, moved to `tsup` for ESM/CJS

carloskelly13:
### Description

- Removes reference to create react app in favor of Vite + NextJS
- We move to `tsup` to build the CJS and ESM outputs, keep `webpack` for UMD, which is required by `one-page`.
- Fixes Spectacle support to work in SSR environments including NextJS App Router
- Adds the `<DefaultTemplate />` component which is often recreated each time a person makes a deck 👇🏻 

default-template

### Updated Docs
**Tutorial** → https://spectacle-docs-website-hsg05kt4y-formidable-labs.vercel.app/docs/tutorial
**Templates** → https://spectacle-docs-website-hsg05kt4y-formidable-labs.vercel.app/docs/api-reference#templates-

Addresses #1260 

#### Type of Change

### How Has This Been Tested?

1. Running the examples including one-page after building the library
2. Following the examples in the documentation and `yarn/npm` linking the library.

*Note* If you are testing one page you will need to build the library and uncomment out the local spectacle dist file as it will be expecting the new `DefaultTemplate` export.


.gitignore
```diff
@@ -44,3 +44,4 @@ bin
 # Pack-ing artifacts
 packages/**/package
 **/*.tgz
+/examples/nextjs-app-router/
```
carloskelly13, line 47: I locally created an example to test against, mixed on if we need to include this as our examples are getting pretty big. But easy to add if we want it committed.

packages/spectacle/src/components/fullscreen.tsx
```diff
@@ -19,6 +19,14 @@ const Container = styled('div')`
 const FullScreen = forwardRef<HTMLDivElement, FSProps>(
   ({ size, color, ...props }, ref) => {
     const toggleFullScreen = useToggleFullScreen();
+    const [isClient, setIsClient] = useState(false);
+
+    useEffect(() => {
+      setIsClient(true);
+    }, []);
```
carloskelly13, line 26: This component requires `document` to exist, so we need to only render it client side.

packages/spectacle/src/hooks/use-slides.tsx
```diff
@@ -49,7 +48,8 @@ export function useSlide(
   doesSlideHaveTemplate: boolean,
   userProvidedId?: SlideId
 ) {
-  const [slideId] = useState<SlideId>(userProvidedId || ulid);
+  const id = useId();
```
carloskelly13, line 51: `ulid` was thrashing on SSR, thus why Spectacle had hydration issues prior. Since any client components are not rendered on the sever (like the fullscreen button) we can guarantee the node tree will remain consistent on first pass and `useId()` is the safer choice.

examples/one-page/index.html
```diff
@@ -83,25 +68,24 @@
...
-          </${Slide}>
-          <${Slide}>
-            <${FlexBox} height="100%" flexDirection="column">
-              <${Heading} margin="0px" fontSize="150px">✨<i>Spectacle</i> ✨</${Heading}>
-              <${Heading} margin="0px" fontSize="h2">A ReactJS Presentation Library</${Heading}>
-              <${Heading} margin="0px 32px" color="primary" fontSize="h3">Where you can write your decks in JSX, Markdown, or MDX!</${Heading}>
-            </${FlexBox}>
-          </${Slide}>
-          <${Slide} transition=${{
+        <${Deck} theme=${theme} template=${html`<${DefaultTemplate} />`}><${Slide}>
```
carloskelly13, line 71: None of this changed other than the `template` prop, but the generation script is keeping the `<${Slide}>` tag same line thus changing all the following indentation. Since this is auto-generated, it's a lower concern.

examples/js/index.js
```diff
@@ -71,6 +52,8 @@ const SlideFragments = () => (
   </>
 );
 
+const template = <DefaultTemplate />;
```
carloskelly13, line 55: This is not inline because the **one page** code is generated from this. The JSX → HTM transpiler does strange things with indentation when you have a render prop like `<A prop={<B />} />` so we just make it a var.

docs/api-reference.md
```diff
@@ -64,6 +64,31 @@ These tags are for displaying textual content.
...
+
+```tsx
+  import { Deck, DefaultTemplate } from 'spectacle'
+
+  const Presentation = () => (
+    <Deck theme={theme} template={<DefaultTemplate />}>
+      {/* Your presentation here */}
+    </Deck>
+  )
+```
```
ryan-roemer, line 82: Minor: dedent block 2 spaces.

packages/spectacle/package.json
```diff
@@ -84,36 +76,27 @@
...
       "files": [
+        "./tsup.config.ts",
         "src/**",
         "!src/**/*.test.*",
         "../../package-scripts.js",
         "../../.babelrc.js",
         "../../.babelrc.build.js"
       ],
       "output": [
-        "es/**/*.js"
```
ryan-roemer, line 109: Minor: You'll want to grep on `["/]es[/]` to catch a few more things still looking for `/es/`

packages/spectacle/package.json
```diff
@@ -84,36 +76,27 @@
...
+      "dependencies": [ "build:lib", "build:dist" ]
     },
     "build:lib": {
-      "dependencies": [
-        "build:lib:esm",
-        "build:lib:cjs"
-      ]
-    },
-    "build:lib:esm": {
-      "command": "nps babel:pkg:lib:esm",
```
ryan-roemer, line 100: Minor: We should clean up `package-scripts.js` for things that are no longer used. For example, I didn't see anything for `babel:pkg:lib`

packages/spectacle/package.json
```diff
@@ -129,37 +112,13 @@
       ]
     },
     "build:dist": {
-      "dependencies": [
-        "build:dist:dev",
```
ryan-roemer, line 133: We don't want to publish a non-minified UMD anymore? 

Side note -- we presently _don't_ publish sourcemaps anywhere -- so usually if we're doing only min we may want to consider publishing those...

package.json
```diff
@@ -119,14 +116,6 @@
         "./examples/typescript:build"
       ]
     },
-    "types:check": {
```
ryan-roemer, line 122: Our contributing guide still has `pnpm run types:check` from root as something you should do. Also, it looks like `types:check` are still available and runnable in our other packages / examples?

packages/spectacle/package.json
carloskelly13: Oh good point. Let me add that back. Do we have any use case for UMD builds outside of one page?

packages/spectacle/src/components/code-pane.tsx
```diff
@@ -12,7 +12,11 @@ import { useSteps } from '../hooks/use-steps';
 import indentNormalizer from '../utils/indent-normalizer';
 import styled, { ThemeContext } from 'styled-components';
 import { compose, layout, position } from 'styled-system';
-import dark from 'react-syntax-highlighter/dist/cjs/styles/prism/vs-dark';
+/**
+ * The types for these theme files don't exist, so we need to ts-ignore it.
+ */
+// @ts-ignore
+import defaultTheme from 'react-syntax-highlighter/dist/esm/styles/prism/vs-dark.js';
```
ryan-roemer, line 19: This breaks Node in CJS mode. Just the following won't work:

```
$ node 
Welcome to Node.js v16.16.0.
Type ".help" for more information.
> const Spectacle = require("./packages/spectacle/lib/index.js")
/Users/rye/scm/fmd/spectacle/node_modules/.pnpm/react-syntax-highlighter@15.5.0_react@18.2.0/node_modules/react-syntax-highlighter/dist/esm/styles/prism/vs-dark.js:3
export default {
^^^^^^

Uncaught SyntaxError: Unexpected token 'export'
```

package.json
carloskelly13: Yeah I can align it to run the same `tsc` command as our other packages. Good call out. `tsup` also does type checking as part of it's process so if you have invalid types, the thing will fail to build.

packages/spectacle/src/components/code-pane.tsx
ryan-roemer: However, as a reminder ESM can import CJS just fine, so maybe go for the CJS version like we had before?

carloskelly13: Shoot. Yeah it needs the extension to work with modules. This one has always been a pain since it's not a real export, but we're just grabbing it from the packages' path.

```
Module not found: Error: Can't resolve 'react-syntax-highlighter/dist/esm/styles/prism/vs-dark' in '/Users/carloskelly13/Developer/spectacle/packages/spectacle/lib'
Did you mean 'vs-dark.js'?
```

without the extension. Maybe we just include this file or make our own theme.

ryan-roemer: Also, now that I'm thinking about it we should add a CI test that with appropriate deps+scripts for wireit effectively does:

```sh
$ cd packages/spectacle
$ pnpm build:lib
$ node -e "require('./lib/index.js')"
```

(This presently fails)

ryan-roemer (commented):
Great start! Mostly minor cleanup, one Node SSR thing that should be straightforward.

Also, presently I'm getting a git diff after install of `pnpm-lock.yaml`

ryan-roemer (commented):
Great start! Mostly minor cleanup, one Node SSR thing that should be straightforward.

Also, presently I'm getting a git diff after install of `pnpm-lock.yaml`

packages/spectacle/package.json
ryan-roemer: Not presently that I know -- and once import maps truly land in browsers then we can finally switch over to just our ESM version.

packages/spectacle/src/components/code-pane.tsx
carloskelly13: I think the solution here it to vendor this theme ourselves since neither import works.

carloskelly13:
I am not getting a thrashing of the pnpm lock file and I think CI would have thrown an error if it had. But I'll check it out again.

packages/spectacle/src/components/code-pane.tsx
ryan-roemer: I would think that adding extension to CJS would work?

```
import dark from 'react-syntax-highlighter/dist/cjs/styles/prism/vs-dark.js';
```

carloskelly13: You're right it does. I was missing the `.default` on the usage since it's not a module. `node -e "require('./lib/index.js')"` runs clean now. Thank you!

packages/spectacle/package.json
carloskelly13: I was struggling to make it work using the `iffe` version `tsup` can build, but we really need some of the polyfills our webpack provides (like `process/browser`) to make this work in the browser.

packages/spectacle/src/hooks/use-slides.tsx
gksander: Do we want to remove ulid as a dependency, and also bump our react peer dep to 18 since we're using useId?

carloskelly13: Yeah! Done. Good call.

gksander (approved):
lgtm

PR #1266, opened 2023-05-09
Export code pane themes object as part of Spectacle.

carloskelly13:
### Description

Currently we make folks jump through installing `react-syntax-highlighter` to use different themes. This just re-exports all the themes and merges in our current default `vsCodeDark` theme.

The size of Spectacle (esm) increases to 124.26 KB from 124.09 KB.

#### Type of Change

### How Has This Been Tested?

Added an example to the TypeScript-demo app.

### DIFF
--- a/packages/spectacle/src/components/code-pane.tsx
+++ b/packages/spectacle/src/components/code-pane.tsx
@@ -13,10 +13,15 @@ import indentNormalizer from '../utils/indent-normalizer';
 import styled, { ThemeContext } from 'styled-components';
 import { compose, layout, position } from 'styled-system';
 /**
- * The types for these theme files don't exist, so we need to ts-ignore it.
+ * We export all the themes from the index file and the VSCode Dark Theme.
+ * The default VSCode dark theme is not part of the index file, so we need
+ * to import it separately and re-export with the rest of the themes.
  */
 // @ts-ignore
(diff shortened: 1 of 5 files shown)


packages/spectacle/src/components/code-pane.tsx
```diff
@@ -13,10 +13,15 @@ import indentNormalizer from '../utils/indent-normalizer';
...
- * The types for these theme files don't exist, so we need to ts-ignore it.
+ * We export all the themes from the index file and the VSCode Dark Theme.
+ * The default VSCode dark theme is not part of the index file, so we need
+ * to import it separately and re-export with the rest of the themes.
  */
 // @ts-ignore
-import defaultTheme from 'react-syntax-highlighter/dist/cjs/styles/prism/vs-dark.js';
+import vsDark from 'react-syntax-highlighter/dist/cjs/styles/prism/vs-dark.js';
+// @ts-ignore
+import * as allThemes from 'react-syntax-highlighter/dist/cjs/styles/prism/index.js';
```
gksander, line 23: Is the separate vsDark import needed here? Seems like it'd be included as part of allThemes, but I might be missing some context.

gksander (approved):
lgtm

packages/spectacle/src/components/code-pane.tsx
carloskelly13: It is not. I put a comment above explaining that so folks down the road aren't confused about this.

PR #1273, opened 2023-05-10
Drop wireit caching from GH Action release build

carloskelly13:
### Description

Having issues with Wireit cache not invalidating on changed files in GH Actions but locally, we want to avoid pushing releases with incomplete assets.


gksander (approved):
dope

PR #1277, opened 2023-05-16
One-page: Pure ESM

ryan-roemer:
Hey! What do you know? `importmap`s now mostly work! Which means we can update our browser one-page version to use just ESM.

Fixes:
- Allow more tolerant `react-syntax-highligher` import behavior to better accommodate both CJS and ESM.

Notes:
- No more UMD needed! 🎉 
- **BUT**, first load if not cached is 2010-era `requirejs`-in-dev-mode slow! 🐢 
- `react*` libs of course don't have ESM versions. And lots of other libraries don't play nice. So instead we use esm.sh which adjusts for ESM things
- We have to manually override react to 18 (since some transitive deps bring in 17)

Future notes:
- [ ] If this works we'll want to ticket out (1) removing UMD and (2) a major release.


ryan-roemer:
**STATUS**: Hold up. Something's rebroken with that prism import...

ryan-roemer:
Oh, haha -- we need to _publish_ my fix first for a preview to work. All good to review @carloskelly13 @gksander !

carloskelly13:
Hitting the local mjs build reveals your theme fix works!

image

examples/one-page/package.json
```diff
@@ -5,6 +5,7 @@
     "pretty": "^2.0.0"
   },
   "scripts": {
+    "start": "pnpm exec serve --listen 5000 ../..",
```
carloskelly13, line 8: This is a nice add over firing up Chrome in insecure mode to bypass CORS.

carloskelly13 (approved):
This is great. We're going to have to keep the React versions inline. I wonder if a future version can fill in a unified React version number for both the import, React DOM, and all the libraries that do a dep on it.

PR #1292, opened 2023-07-26
Add GTAG to Docs website

carloskelly13:
- Added the env vars to our Vercel deployment

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -1,5 +1,7 @@
 // @ts-check
 // Note: type annotations allow type checking and IDEs autocompletion
+require('dotenv').config();
+
 const { romanize } = require('./src/utils/numbers');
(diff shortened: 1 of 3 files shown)


website/docusaurus.config.js
```diff
@@ -1,5 +1,7 @@
 // @ts-check
 // Note: type annotations allow type checking and IDEs autocompletion
+require('dotenv').config();
```
gksander, line 3: GitHub was lagging earlier so you probably didn't see my last commit, but we don't actually need this or the dep on dotenv.

gksander (approved):
Just now seeing this. Looks good. We don't actually need the dotenv dependency (my b on that one)

website/docusaurus.config.js
carloskelly13: Sounds good. I'll nix it!

PR #1304, opened 2023-11-18
Update `create-spectacle` to use import map and ESM-based one page template.

carloskelly13:
The `one-page` output used an outdated version of Spectacle and hard-coded in the dependencies. This implements our dynamic generation for the `one-page.html` option from the `create-spectacle` CLI and moves it over to use ESM and import maps.

How to test → 
1. Pull this PR and cd into `packages/create-spectacle`
2. Either build the library _or_ `pnpm run dev`
3. Pick the options to build a new `one-page` deck
4. Verify the output html file actually opens and loads


scottrippey (approved):

scottrippey (approved):
Instructions worked great

Issue #1323, opened 2024-03-11
spectacle: clear pipeline warnings


Burnett2k:
This project currently has node version warnings in the CICD pipeline logs. This is likely due to using an old version of the below actions. 
        Try and convert to the newest version.
        - actions/checkout → actions/checkout@v4
        - actions/setup-node → actions/setup-node@v4
        - pnpm/action-setup → pnpm/action-setup@v3
        - actions/cache → actions/cache@v4
        - actions/stale → actions/stale@v4 
        In order to conform with Nearform_Commerce OSS standards, please convert to use newer versions which no longer have the warning.

        created by Sawyer Automation Bot © using github cli

Burnett2k:
This also needs to be fixed: `The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/`

Burnett2k:
This is now completed.

PR #1324, opened 2024-03-11
stop using node 16 in pipelines / clear other pipeline warnings

### LINKED ISSUE #1322: spectacle: ensure actions use node18
This project is referencing old versions of Node we no longer want to support (I.e < 18). In order to conform with Nearform_Commerce OSS standards, please convert the action from in matrices and instead run for node 18.

Burnett2k:
### Description

As directed by Cb, we're no longer supporting node16, and will likely not use node matrixes anymore to build / test various node versions.

This pr removes usage of node 16 and matrix, as well as works to clean up some warnings due to using older versions of Github actions such as `setup-node` and `actions/cache` etc.

Fixes #1322 & #1323

### Before updating `ci.yml` (lots of warnings):
image

### After updating `ci.yml` (no warnings):
image

### before: 

image

### after:

image

### before:
image

### after:
release isn't ran until a merge / push is made to master, so I cannot obtain a screenshot of the results without doing some trickery. If needed I can do so, but the changes are essentially the same as the other two files, so seems probably unnecessary.

#### Type of Change

- [x] Chore / tech debt cleanup

### How Has This Been Tested?

1. I verified in the pipelines that the pnpm cache path is still retrieved and set correctly, and warnings were removed
2. I built and ran all the sample applications locally

- [ ] I have included a [changeset](../CONTRIBUTING.md#changesets) if this change will require a version change to one of the packages.
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have run `pnpm run check:ci` and all checks pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,14 +16,11 @@ jobs:
   build:
     name: Check and build codebase
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node-version: [16.x, 18.x]
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: ${{ matrix.node-version }}
+          node-version: 18.x
 
       # Wireit cache
       - uses: google/wireit@setup-github-actions-caching/v1
@@ -52,7 +49,7 @@ jobs:
       # - name: Clear all caches
       #   run: pnpm clean:cache
 
-      - name: Build Code and Examples ${{ matrix.node-version }}
+      - name: Build Code and Examples
         run: pnpm run build
 
       # We build in-source files like `examples/one-page/index.html`.
@@ -60,5 +57,5 @@ jobs:
       - name: Check generated in-source files
         run: git diff --no-ext-diff --quiet --exit-code
 
-      - name: Check Code ${{ matrix.node-version }}
+      - name: Check Code
         run: pnpm run check:ci
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,8 @@
 name: CI
 
+env:
+  PNPM_STORE_PATH: ''
+
 # Runs build and test on:
 #   every push that has a change in a file not in the docs folder
 #   every pull request with main branch as the base that has a change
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -16,30 +19,31 @@ jobs:
...
         with:
+          run_install: false
           version: 7
 
       - name: Get pnpm store directory
+        shell: bash
         id: pnpm-cache
-        run: echo "::set-output name=pnpm_cache_dir::$(pnpm store path)"
+        run: |
+          echo "PNPM_STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_ENV
```
Burnett2k, line 41: this clears the warning outlined [here](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) and in the pipeline logs

carbonrobot: For pnpm the caching is now built into action/setup@v4, you can now remove all the manual cache directives such as lines 3-4, 34, and 37-49.

carbonrobot: For an example, please see https://github.com/FormidableLabs/victory/blob/main/.github/actions/setup/action.yml#L1

Burnett2k: Oh good call. I'm surprised to not see this mentioned anywhere in the pnpm/action-setup documentation. It still lists out the manual strategy for caching.

Burnett2k: @carbonrobot This has been completed. Pipeline logs indicate the cache directory was found 👍🏻

.github/workflows/ci.yml
```diff
@@ -16,33 +16,21 @@ jobs:
...
         with:
-          path: ${{ steps.pnpm-cache.outputs.pnpm_cache_dir }}
-          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('./pnpm-lock.yaml') }}
-          restore-keys: |
-            ${{ runner.os }}-pnpm-store-
+          cache: 'pnpm'
+          node-version: 18.x
+
+      # Wireit cache
+      - uses: google/wireit@setup-github-actions-caching/v1
```
carbonrobot, line 33: Spectacle doesn't currently use [wireit](https://github.com/google/wireit) so we can remove these references.

.github/workflows/create-spectacle.yml
```diff
@@ -24,29 +24,20 @@ jobs:
...
-          node-version: ${{ matrix.node-version }}
-
-      # Wireit cache
-      - uses: google/wireit@setup-github-actions-caching/v1
+      - uses: actions/checkout@v4
 
-      - uses: pnpm/action-setup@v2.2.2
+      - uses: pnpm/action-setup@v3
         with:
+          run_install: false
```
carbonrobot, line 31: `null/false` is the default, so we don't need to specify it here

https://github.com/pnpm/action-setup?tab=readme-ov-file#run_install

.github/workflows/ci.yml
Burnett2k: Are you sure? From what I can tell it's used in most all commands in the root package.json. Maybe I am misunderstanding your comment?

carbonrobot (approved):

PR #1345, opened 2025-01-08
updated docusaurus config with open-graph image and metadata

barret-anspach:
### Description

This PR adds an open-graph image (and metadata) for Spectacle's lander page.

(I've recreated the PR because of pnpm version issues and a series of unsuccessful attempts to restore pnpm-lock to its original condition.)

#### Type of Change

### How Has This Been Tested?

Will test by way of sharing a link to the lander, once this PR is submitted.


website/docusaurus.config.js
```diff
@@ -1,14 +1,16 @@
-// @ts-check
```
nlkluth, line 1: is this no longer needed? if it was doing some type-ish stuff it might be worth keeping it 💭

nlkluth (approved):

website/docusaurus.config.js
barret-anspach: Ah good catch! I didn't mean to remove, will put back in place. Thank you!

PR #1355, opened 2025-09-25
Update OIDC and node versions

ryan-roemer:
- Update release action to use NPM OIDC. This work is modeled on @zibs work in https://github.com/FormidableLabs/victory-native-xl/pull/619/files
    - Note: OIDC connect has been set up in NPM for spectacle, spectacle-mdx-loader, and create-spectacle.
- Update CI to use Node 22.x


.nvmrc
```diff
@@ -1 +1 @@
-18.x
+22.x
```
simoneb, line 1: in the nearform packages, we set this to always be the literal string lts/*, so that we don't have to remember to update things all the time. it automatically picks the latest lts

.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           cache: 'pnpm'
-          node-version: 18.x
+          node-version: 22.x
```
simoneb: since you have the .nvmrc file, this can become very conveniently:


```suggestion
          node-version-file: .nvmrc
```

.github/workflows/create-spectacle.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [18.x]
+        node-version: [22.x]
```
simoneb: same as previous comment, although I'm not sure how it works for matrix, but since we're not using a matrix in the first place, I'd just get rid of it and use the simpler syntax

.github/workflows/release.yml
```diff
@@ -29,7 +31,13 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           cache: 'pnpm'
-          node-version: 18.x
+          node-version: 22.x
```
simoneb: same

.github/workflows/release.yml
```diff
@@ -29,7 +31,13 @@ jobs:
...
+
+      - name: Upgrade npm for OIDC
+        run: npm i -g npm@^11.5.1
```
simoneb: not sure if Eli did this in this way in the other repo(s), but I'd do >= so that major versions are also automatically used. we did it like this in Optic https://github.com/nearform-actions/optic-release-automation-action/blob/5377363b9cb98e3b220cf714d8cdca36d2ff289f/action.yml#L114

.github/workflows/release.yml
```diff
@@ -29,7 +31,13 @@ jobs:
...
+
+      - name: Set npm registry for OIDC
+        run: npm config set registry https://registry.npmjs.org
```
simoneb: I'm not entirely sure if this is actually needed, as I assume it should be the default

.nvmrc
ryan-roemer: I usually leave it to match the minimum engine requirement so that we don't inadvertently introduce new Node features that aren't supported in the allowed engine requirement in a manner that would be missed by CI

.github/workflows/create-spectacle.yml
ryan-roemer: That's a good point!

.github/workflows/release.yml
ryan-roemer: Makes sense!

.github/workflows/release.yml
ryan-roemer: I'll remove it and we can see if we hit any issues.

ryan-roemer:
Great review @simoneb! I think I've covered all the comments now in my latest push.

simoneb (approved):

PR #1366, opened 2026-09-22
build: upgrade minor dependencies

ilteoood:
### Description

Run npm-check-updates across all workspaces with `--target minor` to bump every package to its latest minor (and patch) release.

Notable upgrades:

- React 18.2 → 18.3, @types/react 18.2 → 18.3
- TypeScript 4.8 → 4.9 (root), 5.2 → 5.9 (website)
- eslint 8.24 → 8.57, eslint-plugin-react 7.31 → 7.37
- @docusaurus/* 3.3 → 3.10
- webpack 5.76 → 5.111, webpack-dev-server 4.11 → 4.15
- prettier 2.4 → 2.8
- jest / ts-jest 29.1 → 29.7
- react-spring 9.5 → 9.7, broadcast-channel 4.17 → 4.20
- formidable-oss-badges 1.4 → 1.7
- and ~70 more entries across the root, packages, examples, and the website.

### Changes

Code changes required to keep CI green after the bump:

- `use-full-screen.ts` — TS 4.9 narrows the post-`in` operand type more strictly, so type the webkit-prefixed element explicitly instead of relying on a non-null assertion.
- `slide.tsx` + `print-mode/index.tsx` — TS 4.9 blocks emitting `.d.ts` that reference private `@react-spring/web` paths (TS2742). Annotate `AnimatedDiv` with a public `React.ComponentType<AnimatedProps<any>>` and cast the styled-components selector interpolation accordingly. Also drop the now-unused `@ts-expect-error` on the `onClick` handler.
- `remark-rehype-directive.ts` — prettier 2.8 wraps `typeof X[Y]` in parentheses.
- `examples/mdx/index.js` — `eslint-plugin-react` 7.34+ flags `ReactDOM.render` as deprecated; switch to `createRoot` from `react-dom/client`, matching the other examples.
- `website/tsconfig.json` — TS 5.x enforces `module=Node16` whenever `moduleResolution=Node16` (TS5110), and the upstream `@tsconfig/docusaurus` still references `@docusaurus/theme-classic` (not a direct dep). Override `module` / `moduleResolution` and trim `types` to the package-types alias.
- `website/src/components/index/more-oss.tsx` — `formidable-oss-badges` 1.5+ removed the `featuredBadgeNames` export; build the allow-list locally and render `<FeaturedBadge name=... />` for the known featured products.
- `pnpm-workspace.yaml` — pin `onlyBuiltDependencies` for `@parcel/watcher`, `core-js`, and `esbuild` so pnpm ≥10 stops complaining about ignored postinstall scripts.

### Why

Keep dependencies current for security patches, performance improvements, and compatibility. Several of the upgrades (eslint-plugin-react's deprecation rule, prettier 2.8's formatting change, TS 4.9's stricter `in`-narrowing and declaration-portability rules) would otherwise surface as test/lint/type errors on `main`.

### Test Plan

- `pnpm run check:ci` — passes (lint, prettier, types:check, test across all workspaces).
- `pnpm run build` — passes; `examples/one-page/index.html` is regenerated and included.
- Local install verified with `PUPPETEER_SKIP_DOWNLOAD=true pnpm install`. CI uses pnpm 7 where `pnpm.neverBuiltDependencies` in `package.json` is still respected, so the new `onlyBuiltDependencies` block is additive and won't affect CI.

### Linked Issues

None.


pnpm-workspace.yaml
```diff
@@ -2,3 +2,14 @@ packages:
   - 'packages/*'
   - 'examples/*'
   - 'website'
+
+allowBuilds:
+  '@parcel/watcher': set this to true or false
```
ryan-roemer: I think we need to replace `set this to true or false` with `true` or `false` for all four lines here?

ilteoood: Yep, that's the default output of the pnpm audit.
On it.

ryan-roemer (approved):
I had to add some pnpm config stuff in a commit, but otherwise everything looks good! I tried out all the examples and one-page.

For the future, there's a lot of cruft we'll eventually want to tackle like:
- The whole frontend infra. Get off webpack? (but what does that mean for mdx plugin)
- Get off wireit?

... but for another day. For now, let's just :shipit:

